PR #6, opened 2017-12-22
Change pypi pkg name to 'gordon-dns'

matt0bi:

- Updates pypi package name to `gordon-dns` to avoid name conflict.

N/A

I looked at [namespace packages](https://packaging.python.org/guides/creating-and-discovering-plugins/), but it seemed like a fix for a problem we don't have yet, so I Kept It Simple. 

After this PR is reviewed and merged, I'll go ahead and register the name on pypi  and test-pypi via `python setup.py register`.

**Release note**:

```release-note
NONE
```

cc @spotify/alf 

Update: it looks like travis will always fail for this PR like it does for the `master` and `develop` branches.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,9 +60,10 @@ def install_requires():
 #####
 # Project-specific constants
 #####
-NAME = 'gordon'
+NAME = 'gordon-dns'
+PACKAGE_NAME = 'gordon'
 PACKAGES = find_packages(where='.')
-META_PATH = os.path.join(NAME, '__init__.py')
+META_PATH = os.path.join(PACKAGE_NAME, '__init__.py')
 KEYWORDS = ['dns', 'gcp']
 CLASSIFIERS = [
     'Development Status :: 1 - Planning',


setup.py
```diff
@@ -60,9 +60,10 @@ def install_requires():
 #####
 # Project-specific constants
 #####
-NAME = 'gordon'
+NAME = 'gordon-dns'
+PACKAGE_NAME = 'gordon'
 PACKAGES = find_packages(where='.')
-META_PATH = os.path.join(NAME, '__init__.py')
+META_PATH = os.path.join(PACKAGE_NAME, '__init__.py')
```
econchick: Just so I fully understand this, the `name` arg in `setup.py`:

```py
setup(
    name=NAME,
    # snip
)
```
is therefore `gordon-dns`, and is what defines the name when `pip install`'ing, i.e. `pip install gordon-dns`. Then, the `packages` in `setup`:

```py
setup(
    # snip
    PACKAGES=find_packages(where='.'),
    # snip
)
```
defines the import namespace, as in `import gordon`?

(I'm looking at [PyYaml](https://github.com/yaml/pyyaml/blob/master/setup.py)'s `setup.py` for some guidance.)

RochesterinNYC:
For clarification, the renaming is intended to prevent name conflict with what specifically?

setup.py
matt0bi: Yes - the `name` arg defines what the package will be registered as in pypi while the `packages` arg tells setuptools what "directory level packages" to include in the distribution, so I think that technically, the `gordon` directory name defines the import namespace (`find_packages` just collects it for setuptools).

matt0bi:
@RochesterinNYC There already exists a pypi package called [gordon](https://pypi.python.org/pypi/gordon/0.7.0) so we would not be able to upload another `gordon` package to pypi.

econchick:
@RochesterinNYC "gordon" is already taken on PyPI :-/

RochesterinNYC:
Gotcha, thanks for the clarification.

setup.py
matt0bi: Oh, also, pip installing it locally gives:

```bash
(gordontmp) matt@work:~/code/gordon$ pip list
gordon-dns (0.0.1.dev0, /home/matt/code/gordon)
pip (9.0.1)
setuptools (38.2.4)
wheel (0.30.0)
(gordontmp) matt@work:~/code/gordon$ python
Python 3.7.0a0 (heads/master:122e88a835, Aug 31 2017, 15:49:09) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import gordon
>>>
```

Pip knows it as `gordon-dns`, but `import gordon` works as expected.

PR #9, opened 2018-01-19
Add core interfaces

matt0bi:

It adds a collection of interfaces that define how a plugin and gordon interacts.

N/A

None

**Release note**:

```release-note
NONE
```

cc @spotify/alf

### DIFF
--- a/gordon/interfaces.py
+++ b/gordon/interfaces.py
@@ -1,0 +1,118 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2017 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from zope.interface import Attribute
+from zope.interface import Interface
+
+
+class IEventMessage(Interface):
+    """A discrete unit of work for Gordon to process.
+
+    Gordon expects plugins to return or accept objects that implement this
+    interface in order to route them to other plugins, and handle retries
+    or cleanup in case of errors.
+    """
+
+    def __init__(msg_id, data, history_log, phase=None):
+        """Initialize an EventMessage.
+
+        Args:
+            msg_id (str): Unique message identifier.
+            data (dict): Data required to update DNS records.
+            history_log (list): Log of actions performed on message.
+            phase (str): Current phase.
+        """
+
+    def append_to_history(entry):
+        """Append entry to the IEventMessage's history_log.
+
+        Args:
+            entry (str): Information to append to log.
+        """
+
+
+class IGenericPlugin(Interface):
+    """**Do not** implement this interface directly.
+
+    Use :py:class:`gordon.interfaces.IEventConsumerClient`,
+    :py:class:`gordon.interfaces.IEnricherClient`,
+    or :py:class:`gordon.interfaces.IPublisherClient` instead.
+    """
+    phase = Attribute('Plugin phase')
+
+    def __init__(config, success_channel, error_channel, loop, metrics=None):
+        """Initialize an EventClient object.
+
+        Args:
+            config (dict): plugin-specific configuration
+            success_channel (asyncio.Queue): a sink for successfully processed
+                IEventMessages.
+            error_channel (asyncio.Queue): a sink for IEventMessages that were
+                not processed due to problems.
+            loop (obj): object implementing asyncio.AbstractEventLoop.
+            metrics (obj): Optional obj used to emit metrics.
+        """
+
+    async def update_phase(event_msg):
+        """Update the phase of a message to current phase.
+
+        Args:
+            event_msg (IEventMessage): message with stale phase.
+        """
+
+
+class IEventConsumerClient(IGenericPlugin):
+    """Client for ingesting push/pull events for Gordon to process.
+
+    The client also  receives both successful and failed
+    :py:class:`gordon.interfaces.IEventMessage` objects from Gordon in order
+    to perform cleanup if needed.
+    """
+
+    async def start():
+        """Begin consuming messages using the provided event loop."""
+
+    async def cleanup(event_msg):
+        """Perform cleanup tasks related to a message.
+
+        Args:
+            event_msg (IEventMessage): message at the end of its life cycle.
+        """
+
+
+class IEnricherClient(IGenericPlugin):
+    """Client for processing (enriching) or filtering events received by Gordon.
+
+    Note that if no extra processing is required, the implementer can
+    immediately push the received event into the success_channel.
+    """
+
+    async def process(event_msg):
+        """Process message.
+
+        Args:
+            event_msg (IEventMessage): message to process.
+        """
+
+
+class IPublisherClient(IGenericPlugin):
+    """Client for publishing processed events to their destination."""
+
+    async def publish_changes(event_msg):
+        """Publish processed event to its destination.
+
+        Args:
+            event_msg (IEventMessage): message ready to be sent to destination.
+        """
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,6 +2,7 @@
 branch = True
 source =
     gordon
+omit = gordon/interfaces.py
 
 [paths]
 source =
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,2 +1,3 @@
 -e .
 sphinx==1.6.5
+repoze.sphinx.autointerface==0.8
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -19,3 +19,22 @@ plugins_loader
 .. automodule:: gordon.plugins_loader
 
 .. autofunction:: gordon.plugins_loader.load_plugins
+
+
+interfaces
+----------
+
+.. autointerface:: gordon.interfaces.IEventMessage
+    :members:
+
+.. autointerface:: gordon.interfaces.IEventConsumerClient
+    :members:
+
+.. autointerface:: gordon.interfaces.IEnricherClient
+    :members:
+
+.. autointerface:: gordon.interfaces.IPublisherClient
+    :members:
+
+.. autointerface:: gordon.interfaces.IGenericPlugin
+    :members:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,6 +58,8 @@ extensions = [
     'sphinx.ext.napoleon',
     # link to other projects' documentation
     'sphinx.ext.intersphinx',
+    # allow autodoc to work with zope.interface interfaces
+    'repoze.sphinx.autointerface',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/gordon/__init__.py
+++ b/gordon/__init__.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 __author__ = 'Lynn Root'
-__version__ = '0.0.1.dev0'
+__version__ = '0.0.1.dev1'
 __license__ = 'Apache 2.0'
 __email__ = 'lynn@spotify.com'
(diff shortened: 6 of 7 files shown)


gordon/interfaces.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class IEventMessage(Interface):
+    id = Attribute('Unique message id')
+    phase = Attribute('Current phase of message')
+    data = Attribute('DNS change data')
+    history_log = Attribute('Log of actions performed on the message')
+
+    def __call__(id, data, phase=None):
+        """Initialize an EventMessage.
```
econchick: I can't quite understand from the zope docs why one would choose to use `__call__` instead of `__init__` when declaring an interface. Can you tell me why?

gordon/interfaces.py
```diff
@@ -0,0 +1,86 @@
...
+        Args:
+            config (dict): plugin-specific configuration
+            success_channel (asyncio.Queue): a sink for successfully created
+                IEventMessages.
+            error_channel (asyncio.Queue): a sink for invalid messages.
+            loop (obj): object implementing asyncio.AbstractEventLoop.
+            metrics (obj): Optional obj used to emit metrics.
+        """
+
+    async def update_phase(event):
```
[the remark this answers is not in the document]
econchick: (and same for the rest IMO)

gordon/interfaces.py
```diff
@@ -0,0 +1,86 @@
+from zope.interface import Attribute
+from zope.interface import Interface
+
+
+class IEventMessage(Interface):
```
econchick: Could we get a docstring for the class?

gordon/interfaces.py
```diff
@@ -0,0 +1,86 @@
...
+    id = Attribute('Unique message id')
+    phase = Attribute('Current phase of message')
+    data = Attribute('DNS change data')
+    history_log = Attribute('Log of actions performed on the message')
+
+    def __call__(id, data, phase=None):
+        """Initialize an EventMessage.
+
+        Args:
+            id (int): Unique message identifier.
```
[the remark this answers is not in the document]
matt0bi: Yeah, I think this can wait until v2 so we can tackle this once we get there :). From a quick glance, it looks like different message queues use different id schemes (random, sequential, long, short, etc) so the chance is low, but it's there.

This also made me realize that a message id doesn't have to be an integer, it could be an alphanumeric string, so I changed the type of `msg_id` to `str` (from `int`).

gordon/interfaces.py
matt0bi: Thanks for catching this. I got [this part of the zope docs](https://docs.zope.org/zope.interface/README.html#declaring-provided-interfaces) all wrong. After messing around with interfaces in ipython, I updated all the interfaces to use `__init__`.

The zope docs show an example in which the `__init__` method to be different from other methods because it's a class method, which means that its definition should be part of a `IFooFactory` interface instead of `IFoo`.

I don't see any downsides of letting `IFoo` define an `__init__` method so I'll go with this more concise approach.

gordon/interfaces.py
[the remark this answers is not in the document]
matt0bi: I changed it to the more readable `event_message` :+1:

gordon/interfaces.py
matt0bi: Good idea! I added a docstring, I went for communicating intent and descriptiveness, do you think it explains the role of EventMessages?

matt0bi:
Thanks for checking this out @lynn - I made the requested changes, can you PTAL?

gordon/interfaces.py
```diff
@@ -0,0 +1,107 @@
...
+    phase = Attribute('Plugin phase')
+
+    def __init__(config, success_channel, error_channel, loop, metrics=None):
+        """Initialize an EventClient object.
+
+        Args:
+            config (dict): plugin-specific configuration
+            success_channel (asyncio.Queue): a sink for successfully created
+                IEventMessages.
+            error_channel (asyncio.Queue): a sink for invalid messages.
```
econchick: I think "successfully created IEventMessage" and "invalid messages" is not the correct wording. Not sure what would be correct, but something like "sink for messages when plugin client successfully processes message" and "...failed to process message"

gordon/interfaces.py
```diff
@@ -0,0 +1,107 @@
...
+
+    async def update_phase(event_msg):
+        """Update the phase of a message to current phase.
+
+        Args:
+            event_msg (IEventMessage): message with stale phase.
+        """
+
+
+class IEventClient(IGenericPlugin):
```
econchick: Hmm just thought: we have `IEnricherClient` and `IPublisherClient`. Maybe this should be `IEventConsumerClient`?

gordon/interfaces.py
```diff
@@ -0,0 +1,107 @@
...
+
+    async def start():
+        """Schedule the EventClient task for execution in an event loop.
```
econchick: I am thinking you meant to say `EventClient` and not `IEventClient`, and mean `EventClient` to be the implementing class. But it might be confusing. Perhaps something simpler like "Start consuming (processing?) messages in the provided event loop".

gordon/interfaces.py
```diff
@@ -0,0 +1,107 @@
...
+
+        Args:
+            entry (str): Information to append to log.
+        """
+
+
+class IGenericPlugin(Interface):
+    """Do not implement this interface directly.
+
+    Use IEventClient, IEnricherClient, or IPublisherClient instead.
```
econchick: Could you wrap each object here in back-ticks? [Quick example](https://github.com/spotify/gordon-gcp/blob/develop/gordon_gcp/compute.py#L17). Sphinx will format it nicely :) I think you can even have them auto-linked if you do something like 

```
:ref:`gordon.interfaces.IEventClient`
```

gordon/interfaces.py
```diff
@@ -0,0 +1,107 @@
...
+
+    async def cleanup(event_msg):
+        """Perform cleanup tasks related to a message.
+
+        Args:
+            event_msg (IEventMessage): message at the end of its lifecycle.
+        """
+
+
+class IEnricherClient(IGenericPlugin):
```
econchick: Maybe we can get some docstrings on all three clients like the event message :) I really like the tone/approach you did with the event message one.

gordon/interfaces.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class IEventMessage(Interface):
+    """A discrete unit of work for Gordon to process.
+
+    Gordon expects plugins to return or accept objects that implement this
+    interface in order to route them to other plugins, and handle retries
+    or cleanup in case of errors.
+    """
+    msg_id = Attribute('Unique message id')
```
econchick: another silly zope q: why both an `Attribute` and in the initializer?

matt0bi: I think it makes it clearer that the implementing object is expected to have these attributes vs. just accepting them in the initializer.

gordon/interfaces.py
```diff
@@ -0,0 +1,123 @@
...
+
+    async def update_phase(event_msg):
+        """Update the phase of a message to current phase.
+
+        Args:
+            event_msg (IEventMessage): message with stale phase.
+        """
+
+
+class IEventConsumerClient(IGenericPlugin):
```
econchick: I wonder if we should (or even, if there's a way) declare the phase for each plugin, since we're only supporting certain explicitly defined ones (something like `consume` (or `event`?), `enrich`, `publish`, `done`, `drop`). Or, are we trying to be forward-thinking to allow flexibility?

gordon/interfaces.py
```diff
@@ -0,0 +1,123 @@
...
+    """Client for ingesting push/pull events for Gordon to process.
+
+    The client also  receives both successful and failed
+    :py:class:`gordon.interfaces.IEventMessage` objects from Gordon in order
+    to perform cleanup if needed.
+    """
+
+    async def start():
+        """Begin consuming messages using the provided event loop.
+        """
```
econchick: Nitpick - can this be all on one line?

gordon/interfaces.py
```diff
@@ -0,0 +1,123 @@
...
+        """Process message.
+
+        Args:
+            event_msg (IEventMessage): message to process.
+        """
+
+
+class IPublisherClient(IGenericPlugin):
+    """Client for publishing processed events to their destination.
+    """
```
econchick: Same nitpick here 😬

econchick:
@matt0bi in a separate commit (or PR, which ever) could you also bump the dev version & add git tags so that it's released on PyPI?

gordon/interfaces.py
matt0bi: I'm going to go for the "more flexibility" direction and leave this part more open to change. We can revisit this in a while and see if we need to tighten things up instead.

matt0bi:
@econchick I addressed your feedback, can you please take another look?

Also, I can't reply to the comment about `zope.interface.Attribute` so I'll do it here:

> Having class attributes that are actually expected to be specific to each instantiation of the class doesn't make much sense to me, and will actually lead to ugly behavior.
> So, I'd argue all these attributes should actually just be args to the __init__ method, rather than leading to potential confusion that I had (are these meant to be class attributes?).
> I'm open to being convinced otherwise if you feel strongly about Attribute.

I don't feel strongly about it and I can see how that's confusing. I googled around to see if there's a way to specify class/instance attributes with zope but I haven't found anything. The zope docs do say that methods are _attributes_ on a class, which makes sense, but doesn't really help with specifying object attributes, so went with having the attributes be args to the initializer.

gordon/interfaces.py
```diff
@@ -0,0 +1,118 @@
...
+
+class IEventMessage(Interface):
+    """A discrete unit of work for Gordon to process.
+
+    Gordon expects plugins to return or accept objects that implement this
+    interface in order to route them to other plugins, and handle retries
+    or cleanup in case of errors.
+    """
+
+    def __init__(msg_id, data, phase=None, history_log=None):
```
econchick: Will there ever be a time where a new event message will be initialized with some content in the history log? I'm wondering if there should be some `setter`/getter functionality instead that the `append_to_history` makes use of (I don't know, just typing half-baked/uncaffeinated ideas).

gordon/interfaces.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class IGenericPlugin(Interface):
+    """**Do not** implement this interface directly.
+
+    Use :py:class:`gordon.interfaces.IEventConsumerClient`,
+    :py:class:`gordon.interfaces.IEnricherClient`,
+    or :py:class:`gordon.interfaces.IPublisherClient` instead.
+    """
+    phase = Attribute('Plugin phase')
```
econchick: Just thinking out loud: I wish there was a way - either thru zope or elsewhere - to have an enum of supported values for `phase`. 🤷‍♀️

gordon/interfaces.py
matt0bi: There might be a case when it's init'ed with eg. `[f'Created msg at {datetime.now()}']`, but I think that the usual case will involve init'ing it with an empty list. I'd prefer this to setter/getter methods to keep the interface leaner, unless you feel strongly about this?

gordon/interfaces.py
matt0bi: I think this is possible through zope's [invariants](https://docs.zope.org/zope.interface/README.html#invariants) - it would be good if this invariant was the only place where these phases are defined (eg. `gordon.interfaces.PHASES`). Do you know if we will/already define them some place else, like the config?

gordon/interfaces.py
econchick: Ah, hmm. Well first, "...the usual case will involve init'ing it with an empty list" is not true when you have `history_log=None`. Granted, it shouldn't be `history_log=[]` in the args. But it's not necessarily obvious to a plugin that an event message should be initialed with an empty list. 

Yeah I don't know the best approach, so let's leave it at this and change later if we want.

matt0bi: Good point @econchick, I switched the `history_log` arg into being a required param, which should make it stand out more (and the fact that it should be a mutable sequence).

matt0bi:
Thanks for the patience @econchick! I've addressed your latest feedback, do you mind taking a look?

gordon/interfaces.py
econchick: Constants might be a good idea. The invariant thing might be as well. I'd say let's move on and add/update if we find we need something like this.

econchick:
@matt0bi :+1:

PR #10, opened 2018-03-21
Rename interface methods

econchick:

Essentially the same as [this PR for the janitor interfaces](https://github.com/spotify/gordon-janitor/pull/7) with minor doc updates.

N/A

FYI: I'll probably cut a development release to help avoid [this hack in the gcp plugin](https://github.com/spotify/gordon-gcp/pull/4/commits/e299ddcd82645508f05843abde47e02cb8acd984#diff-b91f3d5bd63fcd17221b267e851608e8R10).

**Release note**:

```release-note
Rename interface methods.
```

@matt0bi @spotify/alf

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,3 +1,3 @@
 -e .
 sphinx==1.6.5
-repoze.sphinx.autointerface==0.8
+sphinxcontrib-zopeext==0.2.1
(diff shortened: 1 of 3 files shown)


econchick:
I'm an idiot - other plugins don't need a `run` method for core to call. After working in the janitor for a while, I forgot that core essentially has a router that consumes from the channels, so there's no reason for plugins to be "started" for consuming.

PR #11, opened 2018-03-30
Clarify branching workflow

matt0bi:

This PR clarifies the workflow around release branches.

N/A

N/A

**Release note**:

```release-note
NONE
```

@econchick @spotify/alf

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -58,6 +58,8 @@ Workflow
 * Do not submit unrelated changes on the same branch/pull request.
 * Multiple commits on a branch/pull request is fine, but all should be atomic, and relevant to the goal of the PR. Code changes for a bug fix, plus additional tests (or fixes to tests) and documentation should all be in one commit.
 * Pull requests should be rebased off of the ``develop`` branch.
+* To finish and merge a release branch, project maintainers should first create a PR to merge the branch into ``develop``. Then, they should merge the release branch into ``master`` locally and push to master afterwards.
+* Bugfixes meant for a specific release branch should be merged into that branch through PRs.
 
 Code
 ****


econchick:
I think this reads fine to me, but since this stems from a convo that you and I had, I'd like at least one other from @spotify/alf to make sure it also makes sense to them.

CONTRIBUTING.rst
```diff
@@ -53,6 +53,8 @@ Workflow
 ********
 
 * This project follows the `gitflow`_ branching model. Please name your branch accordingly.
+   - Project maintainers first create a PR to merge a release branch into the ``develop`` branch, then they merge the release branch into the ``master`` branch locally and push to master.
```
descent3me: This one is clear to me, as long as this is meant as a definition about "what to do when making a release".  I don't have any opinion on the policy, but at least I understand what I'm supposed to do when I read it. Is the reason for adding this line to add emphasis that they should come through a PR to develop?  Otherwise it seems to just directly follow the gitflow model.

CONTRIBUTING.rst
```diff
@@ -53,6 +53,8 @@ Workflow
...
+   - It helps to think of release branches as a staging area. Bugfixes meant for a specific release branch are merged through PRs.
```
descent3me: This one is a little less clear.  To me it seems logical that bug fixes would be made in the same branch as the bugs, so it's confusing that it would need to be said. Is the emphasis on that they should be made through PRs rather than direct commits?  I also don't get a whole lot out of the staging area comment.

CONTRIBUTING.rst
matt0bi: > Is the reason for adding this line to add emphasis that they should come through a PR to develop? 

Yeah, exactly. Afaik, the gitflow model doesn't include platform-specific things like PRs, just basic git operations, so my aim here is to make it clear of what happens on the platform when a release branch is ready to merge into master for both maintainers and contributors.

CONTRIBUTING.rst
matt0bi: > Is the emphasis on that they should be made through PRs rather than direct commits?

Yeah, it's mainly that and partially to emphasize that if someone's submitting a bugfix for a feature that's on a release branch, that fix should be PR'ed to the release branch instead of develop where the bug also exists.

matt0bi:
@descent3me Thanks for taking a look! I addressed your comments.

> (...) which probably isn't as relevant to most contributors as the lines directly below, maybe they could be moved to another section further down?

I think they help set some expectations around the workflow so I'm hesitant to move them to a separate section.

matt0bi:
Hey @descent3me, can you PTAL whenever you have a chance?

CONTRIBUTING.rst
descent3me: It might be good to say that all the way out, like "release branch are merged through PRs to that branch, not to ``develop``." I would also reverse the order of these two statements. Put the directive of what to do first, and then the explanation of why.

matt0bi:
Hey @descent3me, I followed your suggestion and moved the changes to the end of the section. However, I don't think I understand your second comment:

> It might be good to say that all the way out, like "release branch are merged through PRs to that branch, not to develop." I would also reverse the order of these two statements. Put the directive of what to do first, and then the explanation of why.

Could we talk about it face to face sometime today?

descent3me:
sure!

matt0bi:
Hey @descent3me, I applied your advice (thanks!) and changed things around - can you please take a look when you have a moment?

CONTRIBUTING.rst
```diff
@@ -58,6 +58,8 @@ Workflow
 * Do not submit unrelated changes on the same branch/pull request.
 * Multiple commits on a branch/pull request is fine, but all should be atomic, and relevant to the goal of the PR. Code changes for a bug fix, plus additional tests (or fixes to tests) and documentation should all be in one commit.
 * Pull requests should be rebased off of the ``develop`` branch.
+* To finish and merge a release branch, project maintainers should first create a PR to merge the branch into ``develop``. Then, they should merge the release branch into ``master`` locally and push to master afterwards.
+* Bugfixes meant for a specific release branch should be merge into that branch through PRs.
```
descent3me: ...should be *merged*

PR #12, opened 2018-04-20
add async worker to collect hostname resolution metrics

nuritizra:

The ticket intention was to add async worker to collect hostname resolution metrics, but for start I only log the time it takes to check if there is an available record in a given DNS server.
If there isn't an available record we logging a failure msg.

N/A

N/A

**Release note**:
NONE

```release-note

```

@matt0bi @spotify/alf @steven @skinner

### DIFF
--- a/gordon/record_checker.py
+++ b/gordon/record_checker.py
@@ -19,7 +19,6 @@ import logging
 import time
 
 import async_dns
-
 from async_dns.resolver import ProxyResolver
 
 from gordon import exceptions
--- a/gordon/record_checker.py
+++ b/gordon/record_checker.py
@@ -1,3 +1,19 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2017 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import asyncio
 import logging
 import time
@@ -7,7 +23,6 @@ from async_dns.resolver import ProxyResolver
 
 
 class RecordChecker(object):
-    QUERYING_THRESHOLD = 60
 
     def __init__(self, dns_ip):
         """Setup RecordChecker object
@@ -27,14 +42,17 @@ class RecordChecker(object):
         """
         return record.values()
 
-    async def check_record(self, record):
-        """Asynchronously queries a DNS server for a given record
-            to check how long it takes the record to become available.
+    async def check_record(self, record, timeout=60):
+        """Measures the time for a DNS record to become available.
+
+        Query a provided DNS server multiple time until
+            the reply matches the information in the
+            record or until timeout is reached.
 
         Args:
             record (dict): DNS record as a dict
             with record properties
-
+            timeout (int): time threshold to query the DNS server
         """
         start_time = time.time()
 
@@ -42,19 +60,22 @@ class RecordChecker(object):
         r_type_code = types.get_code(r_type)
 
         record_found_in_dns_srv = False
-        query_tries = 0
+        retries = 0
+        sleep_time = 2
 
-        while not record_found_in_dns_srv and\
-                self.QUERYING_THRESHOLD != query_tries:
+        while not record_found_in_dns_srv and \
+                timeout > sleep_time:
 
-            query_tries += 1
+            retries += 1
             resolver_res = await self._resolver.query(name, r_type_code)
             possible_ans = resolver_res.an
 
             record_found_in_dns_srv = \
                 await self._check_resolver_ans(possible_ans, name,
                                                rr_data, ttl, r_type_code)
-            await asyncio.sleep(1)
+
+            sleep_time = self._get_wait_time_exp(retries)
+            await asyncio.sleep(sleep_time)
 
         if not record_found_in_dns_srv:
             logging.info(
@@ -65,7 +86,7 @@ class RecordChecker(object):
 
     async def _check_resolver_ans(
             self, dns_ans_lst, name, rr_data, ttl, r_type_code):
-        """Check if resolver ans is equal to record data.
+        """Check if resolver answer is equal to record data.
 
         Args:
             dns_ans_lst (list): DNS answer list contains record objects
@@ -77,8 +98,9 @@ class RecordChecker(object):
         Returns:
             boolean indicating if DNS ans data is equal to record data
         """
-        type_filtered_lst = \
-            list(filter(lambda r: r.qtype == r_type_code, dns_ans_lst))
+        type_filtered_lst = [
+            ans for ans in dns_ans_lst if ans.qtype == r_type_code
+        ]
 
         # check to see that type_filtered_lst has
         # the same number of records as record rr_data
@@ -98,15 +120,6 @@ class RecordChecker(object):
 
         return True
 
-
-if __name__ == '__main__':
-    loop = asyncio.get_event_loop()
-
-    record_check = RecordChecker('8.8.8.8')
-    my_rec = {
-        "name": "ns1.dnsowl.com",
-        "rrdatas": ["198.251.84.16", "173.254.242.221", "185.34.216.159"],
-        "type": "A",
-        "ttl": 4945
-    }
-    loop.run_until_complete(record_check.check_record(my_rec))
+    def _get_wait_time_exp(self, retries):
+        """Returns the next wait interval, using an exponential back off"""
+        return pow(2, retries)
--- a/gordon/record_checker.py
+++ b/gordon/record_checker.py
@@ -18,8 +18,11 @@ import asyncio
 import logging
 import time
 
-from async_dns import types
+import async_dns
+
 from async_dns.resolver import ProxyResolver
+
+from gordon import exceptions
 
 
 class RecordChecker(object):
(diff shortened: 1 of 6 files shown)


gordon/record_checker.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        start_time = time.time()
+
+        name, rr_data, r_type, ttl = self._extract_record_data(record)
+        r_type_code = types.get_code(r_type)
+
+        record_found_in_dns_srv = False
+        query_tries = 0
+
+        while not record_found_in_dns_srv and\
```
matt0bi: What do you think about making this function wait exponentially before each try (exponential back off)? I'm thinking about the scenario when there are suddenly 30 records to check, and each of these records makes a DNS query every second. Do you think this is a legit concern or premature optimization?

tests/unit/test_record_check.py
```diff
@@ -0,0 +1,153 @@
...
+    assert failed_msg == expected_msg
+
+
+@pytest.mark.asyncio
+async def test_check_number_of_rrdata_is_not_equal(
+        mocker, get_mock_coro,
+        record_checker_instance, caplog,
+        mock_resolve_query, dns_query_response):
+    """Test that we fail on getting number of records
+        from the srv which is not eqaule to the
```
matt0bi: There's a typo in `eqaule` (-> `equal`). I would also avoid using `srv` because there's a DNS record called SRV and this might be confusing and `server` would be this docstring clearer.

requirements.txt
```diff
@@ -3,3 +3,4 @@ setuptools==38.2.5
 toml==0.9.3.1
 ulogger==1.0.1
 zope.interface==4.4.3
+async-dns==1.0.0
```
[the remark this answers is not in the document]
nuritizra: Hi,

the aiodns returns this object:
ares_query_simple_result(host='198.251.84.16', ttl=21511)

which doesn't let me check about the type (I can assume its his type) and without a name...
is that OK?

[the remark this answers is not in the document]
nuritizra: it says: The result type varies depending on the query type... it does it much more complicate that the first library no?

gordon/record_checker.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        self._dns_ip = dns_ip
+        self._resolver = ProxyResolver()
+        self._resolver.set_proxies([self._dns_ip])
+
+    def _extract_record_data(self, record):
+        """Extract record data from a record dict
+
+        Args:
+            record (dict): DNS record as a dict
```
smstone: Can you add a return type as well?

gordon/record_checker.py
```diff
@@ -0,0 +1,112 @@
...
+class RecordChecker(object):
+    QUERYING_THRESHOLD = 60
+
+    def __init__(self, dns_ip):
+        """Setup RecordChecker object
+
+        Args:
+            dns_ip: DNS ip server to query
+        """
+        self._dns_ip = dns_ip
```
smstone: What about the possibility to query multiple upstream resolvers? I could see the case where we want to know how long a record takes to resolve on a set of resolvers, not just one. We do this currently internally with dns-metrics.

gordon/record_checker.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class RecordChecker(object):
+    QUERYING_THRESHOLD = 60
+
+    def __init__(self, dns_ip):
+        """Setup RecordChecker object
+
+        Args:
+            dns_ip: DNS ip server to query
```
smstone: I think this could be better named to something like `dns_server_list` or something if we wanted to support querying of multiple upstreams.

gordon/record_checker.py
smstone: @matt0bi wouldn't introducing an exponential backoff misrepresent the actual resolution time of the record? i could see a case where a large exponential backoff is in place, but the record was resolvable well before we re-attempt. i'd hope the upstream resolvers could handle these metric (once a second) queries fine. either way, do we have an idea of how granular we would want to look at this data or what SLO it would inform?

gordon/record_checker.py
```diff
@@ -0,0 +1,112 @@
...
+            record (dict): DNS record as a dict
+            with record properties
+
+        """
+        start_time = time.time()
+
+        name, rr_data, r_type, ttl = self._extract_record_data(record)
+        r_type_code = types.get_code(r_type)
+
+        record_found_in_dns_srv = False
```
smstone: I read this as a DNS SRV record, but I think mean server? Can I suggest naming this `resolvable_record` or something like this?

gordon/record_checker.py
```diff
@@ -0,0 +1,112 @@
...
+        if not record_found_in_dns_srv:
+            logging.info(
+                f'Sending metric record-checker-failed: {record}')
+        else:
+            final_time = float(time.time() - start_time)
+            logging.info(final_time)
+
+    async def _check_resolver_ans(
+            self, dns_ans_lst, name, rr_data, ttl, r_type_code):
+        """Check if resolver ans is equal to record data.
```
smstone: Can you please use full words here for clarity?

gordon/record_checker.py
```diff
@@ -0,0 +1,112 @@
...
+
+        Returns:
+            boolean indicating if DNS ans data is equal to record data
+        """
+        type_filtered_lst = \
+            list(filter(lambda r: r.qtype == r_type_code, dns_ans_lst))
+
+        # check to see that type_filtered_lst has
+        # the same number of records as record rr_data
+        if len(type_filtered_lst) != len(rr_data):
```
smstone: I'm not sure this check is needed if you are checking below to see if the data matches exactly. Also, this check would be a false positive in a case like this: An old record's data is `[1.1.1.1, 2.2.2.2]` but the new record's data has `[2.2.2.2, 3.3.3.3]`, this check would "pass" just because it's old record and new record has the same length.

tests/unit/test_record_check.py
```diff
@@ -0,0 +1,153 @@
...
+from gordon import record_checker
+
+
+#####
+# Tests & fixtures for running record checker
+#####
+
+@pytest.fixture
+def record_checker_instance():
+    dnc_ip = '8.8.8.8'
```
smstone: dns misspelled here in variable name `dnc_ip`

gordon/record_checker.py
smstone: Also, should we check if the IP is valid before attempting to query it?

tests/unit/test_record_check.py
```diff
@@ -0,0 +1,153 @@
...
+        "rrdatas": ["198.251.84.16", "173.254.242.221", "185.34.216.159"],
+        "type": "A",
+        "ttl": 10686
+    }
+
+    mock_resolve_query.return_value = dns_query_response
+
+    await record_checker_instance.check_record(record_to_check)
+    time_it_took_to_check_record = float(caplog.records[0].msg)
+    assert 0 < time_it_took_to_check_record < 1.005
```
smstone: Can you set the `1.005` in a clear variable name? I believe you're using this number since we are doing an sleep of 1 right?

tests/unit/test_record_check.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+@pytest.mark.asyncio
+async def test_check_record_failed(
```
smstone: It's not immediately clear what you're testing here. I believe you're testing the fact that the record sets do not match up? Given there are several possible failure cases, can you be more explicit what you're testing?

tests/unit/test_record_check.py
```diff
@@ -0,0 +1,153 @@
+# -*- coding: utf-8 -*-
```
smstone: Can you include a test around what happens if `ProxyResolver` raises an exception?

gordon/record_checker.py
matt0bi: This is a good point @smstone. I'm sorry @nuritizra for leading you down a wrong path, I think my suggestion to use exponential back off wasnt right . I talked with @smstone about this over slack we came up with an idea that the check should retry every 5 seconds, which should be enough granularity. But we're not 100% sure what the timeout value should be, because this is tied to how often the publisher plugin will re-POST the request. Maybe we can talk about this tomorrow?

gordon/record_checker.py
nuritizra: so following our conversion, we will leave it to query just one DNS server for now..

gordon/record_checker.py
nuritizra: yes, but its prior test, after that we will check for each ip, so this test covers the case when we get an answer from the dns server for example: 3 record: 1.1.1.1, 2.2.2.2, 3.3.3.3.. and our record has only 2: 1.1.1.1, 2.2.2.2, so the next for loop will return True but this test will return False...

requirements.txt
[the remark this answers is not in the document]
matt0bi: For history's sake: we talked about this IRL and aiodns queries by record type, which means that in our case we would have to perform multiple queries for different records on each check. `async-dns` works more like `dig` in that it returns multiple record types in the response. We're going to stick with `async-dns`.

gordon/record_checker.py
```diff
@@ -0,0 +1,131 @@
...
+        # check to see that type_filtered_lst has
+        # the same number of records as record_ip_list
+        if len(type_filtered_list) != len(record_ip_list):
+            return False
+
+        # check each record data is equal to the given data
+        for rec in type_filtered_list:
+            conditions = [rec.name == record_name,
+                          rec.data in record_ip_list,
+                          rec.ttl == record_ttl]
```
descent3me: Small optimisation could be to this as an if-condition, and-ed together, with the `in` in lookup last.  Since it's the most expensive, this means you would only check it if the other two conditions pass

gordon/record_checker.py
```diff
@@ -0,0 +1,131 @@
...
+
+            if not resolvable_record:
+                await asyncio.sleep(sleep_time)
+
+        if not resolvable_record:
+            logging.info(
+                f'Sending metric record-checker-failed: {record}')
+        else:
+            final_time = float(time.time() - start_time)
+            logging.info(final_time)
```
descent3me: We probably want to add more information to the log message, perhaps some info about the record and/or the server we tried to contact about it?

gordon/record_checker.py
```diff
@@ -0,0 +1,132 @@
...
+        # check to see that type_filtered_lst has
+        # the same number of records as record_ip_list
+        if len(type_filtered_list) != len(record_ip_list):
+            return False
+
+        # check each record data is equal to the given data
+        for rec in type_filtered_list:
+            conditions = [rec.name == record_name,
+                          rec.ttl == record_ttl,
+                          rec.data in record_ip_list]
```
descent3me: I still think it would be better to `and` these together since then they'll be lazily evaluated, rather than evaluating them all first and then calling `all` on the result. @matt0bi What do you think?  I know you had a comment on `all` earlier, and I might be over-optimising since I don't know the use-case for this, what type of records, etc.

matt0bi: Nice catch @descent3me! I do, however, think that using `all()` here makes the condition easier to read (I'm also a fan of using `all()`). From my own looking through DNS data, I think in something like 99.9% of the cases, the number of records in the `record_ip_list` will be pretty small, maybe a handful or two.

descent3me:
Alright, then I'm satsified +1

PR #13, opened 2018-05-07
fix some typos and inconsistencies in formatting

ocsw:
Fixes some typos and inconsistencies in formatting, also removes a NOQA that doesn't seem to be necessary anymore.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,5 +2,6 @@
   - descent3me
   - econchick
   - matt0bi
-  - RockesterinNYC
+  - RochesterinNYC
   - smstone
+  - ocsw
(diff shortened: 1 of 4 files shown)


descent3me:
Everything looks ok, but are the things like capitalization/periods covered in any of the docs/style guide? I tend not to care about them for args lists since those aren't really sentences, so I would actually be more inclined to remove periods there (in the cases where they exist), rather than capitalize, but of course I'll go with whatever standard we chose as long as we're explicit about choosing it.

ocsw:
The caps/periods aren't 100% explicitly listed, but that's how they do it
in the examples, and how we do it elsewhere.

ocsw:
https://google.github.io/styleguide/pyguide.html#Comments

descent3me:
+1

PR #14, opened 2018-05-08
Record checker formatting

ocsw:
More formatting consistency, also rename a variable that says IP but is actually arbitrary DNS data.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/gordon/record_checker.py
+++ b/gordon/record_checker.py
@@ -27,41 +27,40 @@ from gordon import exceptions
 class RecordChecker(object):
 
     def __init__(self, dns_ip):
-        """Setup RecordChecker object
+        """Setup RecordChecker object.
 
         Args:
-            dns_ip: DNS ip server to query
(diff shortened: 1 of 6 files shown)


ocsw:
Honestly, these test shouldn't be testing exact output, it's super fragile.  But anyway should be ok now.

gordon/record_checker.py
```diff
@@ -27,41 +27,40 @@
...
         """
         return record.values()
 
     async def check_record(self, record, timeout=60):
         """Measures the time for a DNS record to become available.
 
-        Query a provided DNS server multiple time until
+        Query a provided DNS server multiple times until
             the reply matches the information in the
             record or until timeout is reached.
```
descent3me: while you're at it, do you want to fix this indentation? i believe it's also incorrect.

ocsw:
Moved to **https://github.com/spotify/gordon/pull/15** because of problems with git.

PR #15, opened 2018-05-08
Record checker cleanup

ocsw:
More formatting consistency, also rename a variable that says IP but is actually arbitrary DNS data.

Fixes #

**Release note**:

```release-note

```


ocsw:
Moved from **https://github.com/spotify/gordon/pull/14** because I had some git problems separating commits that were supposed to be in separate PRs.

tests/unit/test_record_checker.py
```diff
@@ -172,5 +172,5 @@ def mock_resolve_query(mocker, get_mock_coro, record_checker_instance):
     with pytest.raises(exceptions.InvalidDNSHost) as e:
         record_checker.RecordChecker(dns_ip)
 
-    expected_msg = 'RecordChecker got invalid DNS server ip: 8.855.0.5'
+    expected_msg = 'RecordChecker got invalid DNS server IP: 8.855.0.5'
```
descent3me: This is fine, but since you're adding '.'s everywhere else, how about here, too?

descent3me:
Please squash with appropriate message, but otherwise looks good

PR #17, opened 2018-05-22
Update interface definitions + Add logic to start plugins

econchick:

* Update interface definitions for [naming consistency](https://github.com/spotify/gordon-janitor/commit/44f8fa7adc24d8c94aad7189f67ba710482a07a5) & [previously discussed improvements](https://github.com/spotify/gordon-gcp/pull/11#discussion_r181187199). Updates to gordon-gcp interface implementation are required ([an example](https://github.com/spotify/gordon-gcp/blob/develop/src/gordon_gcp/plugins/enricher.py#L244)).
* Add functionality to run installed and activated plugins. This is essentially fork-lifted from [the janitor](https://github.com/spotify/gordon-janitor/commit/adccc9390885e4b48bbb94fd1ebe65dd7fcec902); once this is merged, the janitor's plugin_loader can be removed and this module be imported into its `main` module. Apologies: I did some test cleanup and was too deep into it before I realized that it should be its own commit(s); I was able to separate some out easily into subsequent commits.

N/A

The running of plugins functionality is similar to the [janitor](https://github.com/spotify/gordon-janitor/commit/adccc9390885e4b48bbb94fd1ebe65dd7fcec902#diff-9bc1630266f0f427305e2ff8457ac0e5), but notably different in the following way:

It will run any plugin (whether it implements the 3 interfaces or is a generic one) that has an implemented async `run` method. Our [interface definitions](https://github.com/spotify/gordon/commit/8ed770bbecf9ba05433c9a51abed47b6d380dce6) has only the event consumer defining an async `run` method, so therefore only that one will be started from `main`. The other two will be kicked off once the channel routing logic is implemented (forthcoming).

I did this because I'd like the ability to use generic plugins that don't necessarily implement a prescribed interface. An example: From our product area demo a month or so ago, I created a little HTTP server to do dynamic log level changes while the service is still running (i.e. a `POST` request to an endpoint could update the gordon service logger from `info` to `debug`). After watching a talk at PyCon 2018 on [making deployable applications](https://hynek.me/talks/deploy-friendly/), I got inspiration to make this a proper plugin (with other features, like a liveness and readyness health endpoints, being able to see event messages in flight & dump their history, etc). I'll write a little RFC up to go into more detail, but that's the motivation here.

**Release note**:

```release-note
Add logic to start installed + configured plugins.
```

@spotify/alf

### DIFF
--- a/gordon/interfaces.py
+++ b/gordon/interfaces.py
@@ -35,11 +35,20 @@ class IEventMessage(Interface):
             phase (str): Current phase.
         """
 
-    def append_to_history(entry):
+    def append_to_history(entry, plugin_phase):
         """Append entry to the IEventMessage's history_log.
 
         Args:
             entry (str): Information to append to log.
+            plugin_phase (str): Phase of plugin that created the log
+                entry message.
+        """
+
+    def update_phase(new_phase):
+        """Update the phase of a message to new phase.
+
+        Args:
+            new_phase (str): Phase to update the message to.
         """
 
 
@@ -52,7 +61,7 @@ class IGenericPlugin(Interface):
     """
     phase = Attribute('Plugin phase')
 
-    def __init__(config, success_channel, error_channel, loop, metrics=None):
+    def __init__(config, success_channel, error_channel, metrics=None):
         """Initialize an EventClient object.
 
         Args:
@@ -61,16 +70,11 @@ class IGenericPlugin(Interface):
                 IEventMessages.
             error_channel (asyncio.Queue): A sink for IEventMessages that were
                 not processed due to problems.
-            loop (obj): Object implementing asyncio.AbstractEventLoop.
             metrics (obj): Optional obj used to emit metrics.
         """
 
-    async def update_phase(event_msg):
-        """Update the phase of a message to current phase.
-
-        Args:
-            event_msg (IEventMessage): Message with stale phase.
-        """
+    async def shutdown():
+        """Gracefully shutdown plugin."""
 
 
 class IEventConsumerClient(IGenericPlugin):
@@ -81,7 +85,7 @@ class IEventConsumerClient(IGenericPlugin):
     to perform cleanup if needed.
     """
 
-    async def start():
+    async def run():
         """Begin consuming messages using the provided event loop."""
 
     async def cleanup(event_msg):
--- a/gordon/main.py
+++ b/gordon/main.py
@@ -32,6 +32,7 @@ Example:
     $ python gordon/main.py --config-root /etc/default/
 """
 
+import asyncio
 import logging
 import os
 
@@ -40,6 +41,8 @@ import toml
 import ulogger
 
 from gordon import __version__ as version
+from gordon import exceptions
+from gordon import interfaces
 from gordon import plugins_loader
 
 
@@ -83,18 +86,72 @@ def setup(config_root=''):
     return config
 
 
-def _log_or_exit_on_exceptions(errors, debug):
+def _log_or_exit_on_exceptions(base_msg, exc, debug):
     log_level_func = logging.warn
     if not debug:
         log_level_func = logging.error
 
-    base_msg = 'Plugin "{name}" was not loaded:'
-    for name, exc in errors:
-        msg = base_msg.format(name=name)
-        log_level_func(msg, exc_info=exc)
+    if isinstance(exc, list):
+        for exception in exc:
+            log_level_func(base_msg, exc_info=exception)
+    else:
+        log_level_func(base_msg, exc_info=exc)
 
     if not debug:
         raise SystemExit(1)
+
+
+def _assert_required_plugins(installed_plugins, debug):
+    # enricher not required
+    required_providers_available = {
+        'event_consumer': False,
+        'publisher': False,
+    }
+    for plugin in installed_plugins:
+        if interfaces.IEventConsumerClient.providedBy(plugin):
+            required_providers_available['event_consumer'] = True
+        elif interfaces.IPublisherClient.providedBy(plugin):
+            required_providers_available['publisher'] = True
+
+    missing = []
+    msg = ('The provider for the "{name}" interface is not configured for the '
+           'Gordon service or is not implemented.')
+    for provider, available in required_providers_available.items():
+        if not available:
+            exc = exceptions.MissingPluginError(msg.format(name=provider))
+            missing.append(exc)
+    if missing:
+        base_msg = 'Problem running plugins: '
+        _log_or_exit_on_exceptions(base_msg, missing, debug=debug)
+
+
+def _gather_runnable_plugins(plugins, debug):
+    plugins_to_run = []
+    for plugin in plugins:
+        if interfaces.IEventConsumerClient.providedBy(plugin):
+            # TODO (lynn): this should be switched out for adding
+            # the "verify interface implementation" ability. See
+            # https://docs.zope.org/zope.interface/verify.html
+            if not hasattr(plugin, 'run') or \
+                    not asyncio.iscoroutinefunction(plugin.run):
+                msg = (f'Implemention "{plugin}" of the required '
+                       '"IEventConsumerClient" interface does not have the '
+                       'necessary `run` method.')
+                exc = exceptions.InvalidPluginError(msg)
+                _log_or_exit_on_exceptions(msg, exc, debug)
+                continue
+            plugins_to_run.append(plugin)
+        elif hasattr(plugin, 'run') and asyncio.iscoroutinefunction(plugin.run):
+            plugins_to_run.append(plugin)
+
+    return plugins_to_run
+
+
+async def _run(plugins, debug):
+    _assert_required_plugins(plugins, debug)
+    plugins_to_run = _gather_runnable_plugins(plugins, debug)
+    tasks = [p.run() for p in plugins_to_run]
+    await asyncio.gather(*tasks)
 
 
 @click.command()
@@ -105,16 +162,28 @@ def run(config_root):
     config = setup(os.path.abspath(config_root))
     debug_mode = config.get('core', {}).get('debug', False)
 
-    plugin_names, plugins, errors = plugins_loader.load_plugins(config)
+    # TODO: initialize a metrics object - either here or within `load_plugins`
+    plugin_kwargs = {
+        'success_channel': asyncio.Queue(),
+        'error_channel': asyncio.Queue(),
+    }
+    plugin_names, plugins, errors = plugins_loader.load_plugins(
+        config, plugin_kwargs)
     if errors:
-        _log_or_exit_on_exceptions(errors, debug_mode)
+        for err_plugin, exc in errors:
+            base_msg = 'Plugin was not loaded: {err_plugin}'
+            _log_or_exit_on_exceptions(base_msg, exc, debug=debug_mode)
 
     if plugin_names:
         logging.info(f'Loaded {len(plugin_names)} plugins: {plugin_names}.')
 
     logging.info(f'Starting gordon v{version}...')
-
-    # TODO: actually start the plugins?
+    loop = asyncio.get_event_loop()
+    try:
+        loop.create_task(_run(plugins, debug_mode))
+        loop.run_forever()
+    finally:
+        loop.stop()
 
 
 if __name__ == '__main__':
--- a/gordon/plugins_loader.py
+++ b/gordon/plugins_loader.py
@@ -195,9 +195,10 @@ def load_plugins(config, plugin_kwargs):
         plugin_kwargs (dict): keyword arguments to give to plugins
             during instantiation.
     Returns:
-        Tuple of 2 lists: list of names of plugins, then list of
-        instantiated plugin objects. A tuple of two empty lists is
-        return if there are no plugins found or activated in gordon
+        Tuple of 3 lists: list of names of plugins, list of
+        instantiated plugin objects, and any errors encountered while
+        loading/instantiating plugins. A tuple of three empty lists is
+        returned if there are no plugins found or activated in gordon
         config.
     """
     installed_plugins = _gather_installed_plugins()
--- a/gordon/plugins_loader.py
+++ b/gordon/plugins_loader.py
@@ -64,13 +64,13 @@ from gordon import exceptions
 PLUGIN_NAMESPACE = 'gordon.plugins'
 
 
-def _init_plugins(active_plugins, installed_plugins, plugin_configs):
+def _init_plugins(active, installed, configs, kwargs):
     inited_plugins, inited_plugin_names, errors = [], [], []
-    for plugin_name, unloaded_plugin in installed_plugins.items():
-        if plugin_name not in active_plugins:
+    for plugin_name, unloaded_plugin in installed.items():
+        if plugin_name not in active:
             continue
         plugin_class = unloaded_plugin.load()
-        plugin_config = plugin_configs.get(plugin_name)
+        plugin_config = configs.get(plugin_name)
 
         # check against `None` because `plugin_config` could be `{}`,
         # which should be handled by the plugin's logic (i.e. accept all
@@ -82,7 +82,7 @@ def _init_plugins(active_plugins, installed_plugins, plugin_configs):
             continue
 
         try:
-            plugin_object = plugin_class(plugin_config)
+            plugin_object = plugin_class(plugin_config, **kwargs)
             inited_plugins.append(plugin_object)
             inited_plugin_names.append(plugin_name)
         except Exception as e:
@@ -186,12 +186,14 @@ def _gather_installed_plugins():
     return gathered_plugins
 
 
-def load_plugins(config):
+def load_plugins(config, plugin_kwargs):
     """
     Discover and instantiate plugins.
 
     Args:
         config (dict): loaded configuration for the Gordon service.
+        plugin_kwargs (dict): keyword arguments to give to plugins
+            during instantiation.
     Returns:
         Tuple of 2 lists: list of names of plugins, then list of
         instantiated plugin objects. A tuple of two empty lists is
@@ -204,4 +206,5 @@ def load_plugins(config):
         return [], [], []
     plugin_namespaces = _get_plugin_config_keys(active_plugins)
     plugin_configs = _load_plugin_configs(plugin_namespaces, config)
-    return _init_plugins(active_plugins, installed_plugins, plugin_configs)
+    return _init_plugins(
+        active_plugins, installed_plugins, plugin_configs, plugin_kwargs)
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -17,13 +17,76 @@
 Module for reusable pytest fixtures.
 """
 
+import asyncio
+import logging
+
 import pkg_resources
 import pytest
+import zope.interface
+
+from gordon import interfaces
 
 
-class FakePlugin:
-    def __init__(self, config):
+@zope.interface.implementer(interfaces.IEventConsumerClient)
+class EventConsumerStub:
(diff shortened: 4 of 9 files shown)


gordon/main.py
```diff
@@ -105,14 +164,28 @@ def run(config_root):
...
 
     logging.info(f'Starting gordon v{version}...')
+    loop = asyncio.get_event_loop()
+    try:
+        loop.create_task(_run(plugins, debug_mode))
+        loop.run_forever()
+    finally:
+        loop.stop()
 
     # TODO: actually start the plugins?
```
ocsw: this can be removed

gordon/interfaces.py
```diff
@@ -52,43 +60,38 @@ class IGenericPlugin(Interface):
...
     phase = Attribute('Plugin phase')
 
-    def __init__(config, success_channel, error_channel, loop, metrics=None):
+    def __init__(config, success_channel, error_channel, metrics=None):
         """Initialize an EventClient object.
 
         Args:
-            config (dict): Plugin-specific configuration.
-            success_channel (asyncio.Queue): A sink for successfully processed
+            config (dict): plugin-specific configuration
```
ocsw: Why these changes?  The style guide has the capital and period.

gordon/interfaces.py
```diff
@@ -52,43 +60,38 @@ class IGenericPlugin(Interface):
...
-        """
+    async def shutdown():
+        """Gracefully shutdown plugin."""
 
 
 class IEventConsumerClient(IGenericPlugin):
     """Client for ingesting push/pull events for Gordon to process.
 
-    The client also receives both successful and failed
+    The client also  receives both successful and failed
```
ocsw: extra space

gordon/main.py
```diff
@@ -105,14 +164,28 @@ def run(config_root):
...
+    plugin_kwargs = {
+        'success_channel': asyncio.Queue(),
+        'error_channel': asyncio.Queue(),
+    }
+    plugin_names, plugins, errors = plugins_loader.load_plugins(
+        config, plugin_kwargs)
     if errors:
-        _log_or_exit_on_exceptions(errors, debug_mode)
+        for err_plugin, exc in errors:
+            base_msg = 'Plugin was not loaded: {err}'
```
ocsw: Was this a typo?  I don't see err defined here.

gordon/main.py
```diff
@@ -40,6 +41,8 @@
 import ulogger
 
 from gordon import __version__ as version
```
ocsw: Should this be imported less directly (per the style guide?

gordon/main.py
```diff
@@ -83,20 +86,76 @@ def setup(config_root=''):
...
+            if not hasattr(plugin, 'run') or \
+                    not asyncio.iscoroutinefunction(plugin.run):
+                msg = (f'Implemention "{plugin}" of the required '
+                       '"IEventConsumerClient" interface does not have the '
+                       'necessary `run` method.')
+                exc = exceptions.InvalidPluginError(msg)
+                _log_or_exit_on_exceptions(msg, exc, debug)
+                continue
+            plugins_to_run.append(plugin)
+        elif hasattr(plugin, 'run'):
```
ocsw: Should this also test for async?

gordon/main.py
```diff
@@ -83,20 +86,76 @@ def setup(config_root=''):
...
+    for plugin in plugins:
+        if interfaces.IEventConsumerClient.providedBy(plugin):
+            required_providers['event_consumer'] = True
+        elif interfaces.IEnricherClient.providedBy(plugin):
+            required_providers['enricher'] = True
+        elif interfaces.IPublisherClient.providedBy(plugin):
+            required_providers['publisher'] = True
+
+    missing = []
+    msg = ('The provider for "{name}" interface is not configured for the '
```
ocsw: for "{name}" -> for the "{name}"

gordon/plugins_loader.py
```diff
@@ -64,13 +64,13 @@
...
 
-def _init_plugins(active_plugins, installed_plugins, plugin_configs):
+def _init_plugins(active, installed, configs, kwargs):
     inited_plugins, inited_plugin_names, errors = [], [], []
-    for plugin_name, unloaded_plugin in installed_plugins.items():
-        if plugin_name not in active_plugins:
+    for plugin_name, unloaded_plugin in installed.items():
+        if plugin_name not in active:
             continue
         plugin_class = unloaded_plugin.load()
```
ocsw: Not in the diff, but what does load() do and where is it defined?  It's not in this repo.

gordon/main.py
econchick: While this isn't a part of this PR, importing meta stuff from `__init__` is a special case IMO. `import gordon; gordon.__version__` would work, but could break if we do anything more fancy with `gordon.__init__` (e.g. similar to how we use `gordon_gcp.__init__` and `gordon_janitor_gcp.__init__`, not that we would) since it'd most likely introduce a circular dependency.

gordon/main.py
econchick: Maybe not _test_, but yea I think adding a condition of `asyncio.iscoroutinefunction` is reasonable.

gordon/plugins_loader.py
ocsw: (Maybe add a comment)

gordon/plugins_loader.py
```diff
@@ -82,7 +82,7 @@ def _init_plugins(active_plugins, installed_plugins, plugin_configs):
             continue
 
         try:
-            plugin_object = plugin_class(plugin_config)
+            plugin_object = plugin_class(plugin_config, **kwargs)
             inited_plugins.append(plugin_object)
             inited_plugin_names.append(plugin_name)
         except Exception as e:
```
ocsw: Can this be more specific?

gordon/plugins_loader.py
econchick: The `unloaded_plugin` is actually an instance of pkg_resources's [`EntryPoint`](http://setuptools.readthedocs.io/en/latest/pkg_resources.html#entrypoint-objects)s.

tests/unit/test_plugins_loader.py
```diff
@@ -97,160 +119,156 @@ def test_init_plugins_exceptions(mocker):
...
     """Skips plugins that are not activated in core config."""
-    active_plugins = ['one.plugin']
     inited_names, inited_plugins, errors = plugins_loader._init_plugins(
-        active_plugins, plugins, plugin_config)
+        [conftest.REGISTERED_ACTIVE_PLUGINS[0]], installed_plugins,
+        plugin_config, plugin_kwargs)
 
     assert 1 == len(inited_plugins) == len(inited_names)
-    assert plugin_config.get('one.plugin') == inited_plugins[0].config
+    exp = plugin_config.get(conftest.REGISTERED_ACTIVE_PLUGINS[0])
```
ocsw: Exp was a little unclear to me at first.  Maybe expected?

gordon/main.py
econchick: Actually I do already have this check [here](https://github.com/spotify/gordon/pull/17/commits/d427d4dfe735b5c79590fb32af68879a7ea5495c#diff-6ef2c41f2a81ebf1e2f3f7c8065b9160R138).

gordon/plugins_loader.py
econchick: Not really, we do want to catch any exception that loading a plugin might throw.

gordon/main.py
ocsw: That's for plugins that implement an interface.  What about other runnables?  Can you do an await on them in run() if they aren't async?

tests/unit/conftest.py
```diff
@@ -36,54 +107,94 @@ class FakePluginException(Exception):
...
+    return {
+        'event_consumer': EventConsumerStub,
+        'enricher': EnricherStub,
+        'publisher': PublisherStub,
+        'runnable': GenericStub
+    }[plugin_name]
+
+
+def _get_plugin_conf(plugin_name):
+    plugin_name = plugin_name.split('.')[0]
```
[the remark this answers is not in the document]
econchick: It's being called from [here](https://github.com/spotify/gordon/pull/17/commits/d427d4dfe735b5c79590fb32af68879a7ea5495c#diff-c2d555a913402fd4b764d59ea63a2266R206), but it is cruft left over from when I was trying to clean up tests and didn't separate out stuff nicely. Will remove.

tests/unit/test_main.py
```diff
@@ -99,46 +103,126 @@ def test_log_or_exit_on_exceptions_debug(plugin_exc_mock, mocker, monkeypatch):
...
     logging_mock.warn.assert_called_once()
     logging_mock.error.assert_not_called()
 
 
-run_args = 'has_active_plugins,exp_log_count'
-run_params = [
-    (True, 2),
-    (False, 1),
-]
+@pytest.fixture
```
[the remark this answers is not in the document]
econchick: There's a convo a few months back w/ @descent3me and maybe @matt0bi that it makes more sense to place fixtures closer to where they're used. There's some inconsistencies, but I've tried to adhere to that going forward.

gordon/main.py
econchick: ah right missed that part. I was like, I swore I already typed `asyncio.iscoroutinefunction` somewhere here..

econchick:
@ocsw thank you for the review - I think I addressed most, if not everything. Mind re-reviewing when you get a chance?

gordon/main.py
ocsw: bump

gordon/main.py
ocsw: Can you add a test for this case?  The existing one only hits the interface cases.

gordon/plugins_loader.py
ocsw: Is there anything we should reraise, then?

ocsw:
There were 3 things I'd go back to.  The one I bumped, the one about the async test, and the one about catching all exceptions.  I commented on them.

tests/unit/conftest.py
```diff
@@ -28,6 +28,13 @@
 from gordon import interfaces
 
 
+PLUGIN_NAMES = ['event_consumer', 'enricher', 'publisher', 'runnable']
```
smstone: where does `runnable` plugin come from?

gordon/main.py
econchick: damnit I swore I got this one. maybe didn't commit it.

gordon/plugins_loader.py
econchick: Depends on if the user sets the `debug = True` in config. It's handled in `main:_log_or_exit_on_exceptions`

econchick:
@ocsw I addressed the first and last, but what's the one about the async test you're referring to?

gordon/interfaces.py
```diff
@@ -35,11 +35,19 @@ def __init__(msg_id, data, history_log, phase=None):
...
+    def append_to_history(entry, plugin):
         """Append entry to the IEventMessage's history_log.
 
         Args:
             entry (str): Information to append to log.
+            plugin (str): Plugin that created the log entry message.
+        """
+
+    def update_phase(new_phase):
+        """Update the phase of a message to current phase.
```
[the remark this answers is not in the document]
econchick: good catch, thanks

tests/unit/conftest.py
```diff
@@ -37,53 +100,109 @@ class FakePluginException(Exception):
...
+        plugin_mock = mocker.MagicMock(pkg_resources.EntryPoint)
         plugin_mock.name = name
-        plugin_mock.load.return_value = FakePlugin
-        plugins[name] = plugin_mock
+        plugin_mock.load.return_value = _get_stub_class(name)
+        return plugin_mock
+
+    return {
+        'abc.event_consumer': mock_plugin('abc.event_consumer'),
+        'abc.enricher': mock_plugin('abc.enricher'),
```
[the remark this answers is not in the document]
smstone: `xyz` works for me

tests/unit/conftest.py
econchick: The context is explained in the PR message itself (LMK if it doesn't make sense). This name refers to the [ability to run plugins that aren't necessarily interface implementations](https://github.com/spotify/gordon/pull/17/commits/2733fc4e7bcdb198154344fd2c23b8418b4d6c61#diff-6ef2c41f2a81ebf1e2f3f7c8065b9160R130).

Does that help?

tests/unit/test_main.py
```diff
@@ -99,46 +103,126 @@ def test_log_or_exit_on_exceptions_debug(plugin_exc_mock, mocker, monkeypatch):
...
+def mock_log_or_exit_on_exc(mocker, monkeypatch):
+    mock = mocker.Mock()
+    monkeypatch.setattr('gordon.main._log_or_exit_on_exceptions', mock)
+    return mock
+
+
+@pytest.mark.parametrize('plugins,exp_log_calls', (
+    (('event_consumer', 'enricher', 'publisher', 'runnable'), 0),
+    (('event_consumer', 'enricher', 'publisher'), 0),
+    (('event_consumer', 'enricher'), 1),
```
ocsw: We probably don't need this case if the enricher becomes optional.

tests/unit/test_main.py
```diff
@@ -99,46 +103,126 @@ def test_log_or_exit_on_exceptions_debug(plugin_exc_mock, mocker, monkeypatch):
...
+@pytest.mark.parametrize('has_run,has_run_async,exp_plugins,exp_mock_call', (
+    (True, True, 2, 0),  # has `async def run` method
+    (True, False, 1, 1),  # has `def run` method - not async
+    (False, False, 1, 1),  # does not have a any `run` method
+))
+def test_gather_runnable_plugins(has_run, has_run_async, exp_plugins,
+                                 exp_mock_call, monkeypatch,
+                                 inited_plugins, mock_log_or_exit_on_exc):
+    """Return all plugins with an async `run` method."""
+    patch_run = 'tests.unit.conftest.EventConsumerStub.run'
```
ocsw: This is the bit I meant about async.  We should also test that it does nothing if a runnable's run() isn't async.  Or isn't present, for that matter.

ocsw: Also, if the runnable has a run() but it's not async, should we surface that as a warning or something?

tests/unit/test_main.py
econchick: It's fine to have it - it's the same expected result, and no harm in an extra test.

tests/unit/test_main.py
econchick: Maybe I'm not understanding correctly, but isn't this case tested in the [second iteration](https://github.com/spotify/gordon/pull/17/commits/2733fc4e7bcdb198154344fd2c23b8418b4d6c61#diff-a324e09b40482c4a38ed3c5fd93c0f4eR138) of this test? There's only one runnable plugin when four are available, two having a run method, and only one of those two actually async.

A warning on a sync run is a good suggestion.

econchick: Actually, re: warning - we only get into this case because we don't yet verify interface implementations, which is a v2 thing.

econchick:
@ocsw @smstone hey folks - I think I addressed the remaining suggestions. But I might have missed some between yesterday afternoon and this AM 👼

tests/unit/test_main.py
```diff
@@ -156,7 +156,7 @@ def test_gather_runnable_plugins(has_run, has_run_async, exp_plugins,
 
 
 @pytest.mark.asyncio
-async def test_run_plugins(inited_plugins, mocker, monkeypatch):
```
matt0bi: :bowing_man:

gordon/main.py
```diff
@@ -83,20 +86,74 @@ def setup(config_root=''):
...
+    else:
+        log_level_func(base_msg, exc_info=exc)
 
     if not debug:
         raise SystemExit(1)
 
 
+def _assert_required_plugins(plugins, debug):
+    # enricher not required
+    required_providers = {
```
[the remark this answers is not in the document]
descent3me: Alternatively, I think this might actually be clearer if each plugin was tested individually, rather than added to a dict which gets looped through. I might think if I wanted another plugin to be required, all I had to do was add it here, but actually I would have to add a specific if-block in the loop below as well.  Since each item ends up having a specific if-block anyways, I don't see the value of abstracting it away to a new var.

tests/unit/test_main.py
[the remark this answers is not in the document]
descent3me: Yep, I definitely think it makes it easier to understand a test, if I can see the mocks and the data that is being tested without having to scroll around, check other files, etc.

gordon/main.py
```diff
@@ -83,20 +86,76 @@ def setup(config_root=''):
...
+    if missing:
+        base_msg = 'Issue running plugins: '
+        _log_or_exit_on_exceptions(base_msg, missing, debug=debug)
+
+
+def _gather_runnable_plugins(plugins, debug):
+    plugins_to_run = []
+    for plugin in plugins:
+        if interfaces.IEventConsumerClient.providedBy(plugin):
+            # TODO (lynn): this should be switched out for adding
```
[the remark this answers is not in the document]
econchick: I totally agree I think it's time to introduce it, but I feel like it's out of scope for the ticket I'm working on, unless you feel strongly otherwise.

gordon/main.py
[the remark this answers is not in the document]
econchick: Ah interesting, didn't even occur to me. Thanks for pointing that out!

gordon/main.py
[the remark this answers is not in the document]
econchick: To expand: adding verification requires work to update the GCP plugin as well. So yeah I think verification should be its own task separate from this.

gordon/main.py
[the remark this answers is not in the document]
econchick: For the lurkers, @descent3me and I talked about simplifying this part of the code and it doesn't seem to be possible. I'm renaming a couple of variables to hopefully improve readability.

gordon/main.py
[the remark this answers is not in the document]
matt0bi: ^ Gotcha, makes sense.

tests/unit/test_main.py
ocsw: We're only patching one of the interfaced plugins.  I was suggesting having tests on the non-interface "runnable" ones that we identify based just on their having an async run().

gordon/main.py
```diff
@@ -83,20 +86,74 @@ def setup(config_root=''):
...
+            if not hasattr(plugin, 'run') or \
+                    not asyncio.iscoroutinefunction(plugin.run):
+                msg = (f'Implemention "{plugin}" of the required '
+                       '"IEventConsumerClient" interface does not have the '
+                       'necessary `run` method.')
+                exc = exceptions.InvalidPluginError(msg)
+                _log_or_exit_on_exceptions(msg, exc, debug)
+                continue
+            plugins_to_run.append(plugin)
+        elif hasattr(plugin, 'run') and asyncio.iscoroutinefunction(plugin.run):
```
ocsw: This was the spot I had in mind for the warning - if a plugin isn't implementing an interface, and has run(), but it's not async, it would be helpful to know that for debuggin.

ocsw:
I'm good except for the two async bits (code and test).  But should the commits also be squashed?

gordon/main.py
econchick: Hmm I get what you mean but I don't feel comfortable assuming that a synchronous `run` for an implementation wasn't on purpose.

econchick:
@ocsw we try to maintain [atomic commits](https://www.freshconsulting.com/atomic-commits/). I would squash commits if I were committing changes to the patch as a result of a review, but I do that locally to keep commit history clean. The changes here are different enough tasks/purposes to warrant separate commits.

ocsw:
+1

PR #18, opened 2018-05-29
Add initial routing logic for event messages

econchick:
Right now we just drop all messages that error out. The next commit(s) will add functionality to retry messages.

Add initial routing logic for event messages.

N/A

This does not yet allow for retries; it will drop any event message that errors out. Retry logic will be added after the initial release of Gordon. However if retry logic is needed, then it can be handled within the plugins individually.

**Release note**:

```release-note
Add initial routing logic for event messages.
```

@spotify/alf

### DIFF
--- a/gordon/router.py
+++ b/gordon/router.py
@@ -1,0 +1,170 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2018 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Core message routing logic for the plugins within Gordon Service.
+
+Messages received on the success channel will be routed to the next
+designated plugin phase. For example, a message that has a ``consume``
+phase will be routed to the installed enricher provider (or publisher
+provider if no enricher provider is installed).
+
+If a message fails its next phase, its phase will be updated to ``drop``
+and routed to the event consumer provider for cleanup.
+
+.. attention::
+
+    The :class:`GordonRouter` only supports the following two phase
+    routes:
+
+    1. consume -> enrich -> publish -> done
+    2. consume -> publish -> done
+
+    Future releases may support more configurable phase routings.
+
+"""
+
+import asyncio
+import logging
+
+from gordon import interfaces
+
+
+# TODO/NOTE (lynn): This is a temporary helper function to alias
+#                   entrypoints of plugins to a common name. In the
+#                   future, plugin interfaces will be updated to have a
+#                   common name for their entrypoint.
+def _set_plugin_entrypoint_aliases(plugin_dict):
+    # e.g. enricher_inst.process => enricher_inst.handle_message
+    entrypoints = {
+        'enricher': 'process',
+        'publisher': 'publish_changes',
+        'event_consumer': 'cleanup',
+    }
+    for plugin_name, func in entrypoints.items():
+        try:
+            plugin = plugin_dict[plugin_name]
+        except KeyError:  # if no enricher
+            continue
+        entry_point = getattr(plugin, func)
+        setattr(plugin, 'handle_message', entry_point)
+    return plugin_dict
+
+
+class GordonRouter:
+    """Route messages to the appropriate plugin destination.
+
+    .. attention::
+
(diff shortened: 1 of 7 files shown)


tests/unit/test_router.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+@pytest.fixture()
+def router_inst(inited_plugins, plugin_kwargs):
+    inited_plugins.pop('runnable')
+    router_inst = router.GordonRouter(plugins=inited_plugins, **plugin_kwargs)
+    return router_inst
+
+
+class EventMsgStub:
```
descent3me: Is there a way [/any reason not] to explicitly use the interface in the class definition?  I realize I am missing some context on zope, and anyways we're not using it yet, but I thought we had mentioned somewhere that there is a way to enforce or assert that all required methods are implemented?

The thing I'm concerned about is if in the future we change the IEventMessage, but forget to update some routing method, our tests won't catch it because they're only testing against a Stub that maybe has nothing to do with the current state of the interface anymore.

tests/unit/test_router.py
```diff
@@ -0,0 +1,158 @@
...
+async def test_poll_channel(enricher, raises, router_inst, event_msg, caplog,
+                            monkeypatch):
+    next_phase = 'enrich'
+    if not enricher:
+        router_inst.phase_route.pop('enrich')
+        router_inst.phase_route['consume'] = next_phase = 'publish'
+
+    if raises:
+        next_phase = 'drop'
+        event_msg.phase = 'its_not_a_phase_mom'
```
descent3me: lol

gordon/router.py
```diff
@@ -0,0 +1,158 @@
...
+            processed due to problems.
+        plugins (dict): Implemented plugin provider names mapped to
+            their instantiated objects.
+    """
+    # TODO (lynn): Ideally this is configurable/extendable in future
+    #              iterations of gordon.
+    PHASE_TO_PLUGIN_MAPPER = {
+        'enrich': 'enricher',
+        'publish': 'publisher',
+        'done': 'event_consumer',
```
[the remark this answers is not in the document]
econchick: How about `cleanup`? since that's the method name attached to `event_consumer` that will be called?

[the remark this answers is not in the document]
descent3me: I feel like that might be too closely tied, since messages in 'drop' will also go to cleanup right?  I'm assuming the reason 'drop' and 'done' are separate here is so its obvious for implementers that they *can* do different things.  If one is named the same as the method they both go to, it might be more confusing why 'drop' is there as a separate phase.

tests/unit/test_router.py
econchick: Are you meaning something [like this](https://github.com/spotify/gordon/blob/develop/tests/unit/conftest.py#L38)? If so, then I just forgot to add that decorator to this. But it's also not _really_ needed because I don't do anything w/i the router that checks to see if the event message provided is in fact a provider for the interface. Typing that out makes me feel like I should somewhere, though... and agreeing with what you're saying w/r/t testing. I'll add that.

FWIW The functionality to assert a correct implementation isn't in gordon (or gordon janitor) yet (there's an internal ticket for that). The functionality to find out who _says_ they've implemented an interface exists in gordon/main.py - but anyone can declare they implement something with that aforementioned decorator. Not sure if that makes sense.

econchick:
@descent3me hej hej! I think I addressed everything; lmk if naming is still not that great. PTAL when you get a chance! 🙏 🌮

gordon/router.py
```diff
@@ -0,0 +1,162 @@
...
+    def _setup_phase_route(self):
+        path = {
+            'consume': 'publish',
+            'publish': 'cleanup',
+            'cleanup': 'cleanup',
+        }
+        if 'enricher' in self.plugins:
+            path['consume'] = 'enrich'
+            path['enrich'] = 'publish'
+        return path
```
descent3me: Tell me if I'm being too picky since it's just an local var, but could we call it `route` as well?

tests/unit/test_router.py
descent3me: Yep, that was the thing I meant!  I realize it doesn't do much yet, so its not a blocker for me to skip it.

econchick: Hey I just pushed another commit - I'll squash this if it looks good to you - that basically ignores event messages that do not "provide" the `IEventMessage` interface (to be switched to "verify" once we work on that ability).

gordon/router.py
```diff
@@ -120,6 +122,12 @@ def _get_next_phase(self, event_msg):
         return next_phase
 
     async def _route(self, event_msg, next_phase=None):
+        if not interfaces.IEventMessage.providedBy(event_msg):
+            msg = (f'Ignoring message "{event_msg}". Does not correctly '
+                   'implement `IEventInterface`.')
```
descent3me: I think you mean `IEventMessage`

econchick: ffffff you're right

econchick: @descent3me fixed! I feel like a dummy 🤦‍♀️

PR #19, opened 2018-05-29
Initial interface definitions for metrics.

econchick:

Defines interfaces for metrics plugins.

N/A

In discussing with @matt0bi, we kept this as generic as possible for now. It shouldn't be too difficult for someone to write - for instance - a [prometheus plugin](https://prometheus.io/docs/instrumenting/writing_clientlibs/) with their [python client](https://github.com/prometheus/client_python).

What @matt0bi and I imagine is within gordon core, it would load the metrics provider (very similar to how I approached it [here](https://github.com/spotify/gordon/pull/18/files#diff-6ef2c41f2a81ebf1e2f3f7c8065b9160R151)). If it doesn't find any, it would look to a `metrics.py` module (that's inside of core) to provide a metrics object.  Within `metrics.py`, it will try to import [`aioshumway`](https://github.com/spotify/aioshumway/blob/master/aioshumway/__init__.py#L86), and if it can't, it will default to a object that just logs out any metrics given. 

This allows us to not have to separately package an interface implementation for aioshumway. We'd add `extras_require={'aioshumway': ['aioshumway']}` and then `pip install gordon['aioshumway']` similar to how to do with [ulogger](https://github.com/spotify/ulogger/blob/master/setup.py#L103).

The intended API of the implemented `aioshumway` interface (and default metric logger interface) would be pretty much exactly how we use shumway already:

```python
metrics = aioshumway.MetricRelay('gordon')  # or a logger with the same API
# initialize available plugins w/ metrics object during loading, similar to the channels

# implements publisher interface
class SomePublisher:
    def __init__(self, config, ..., metrics):
        self.config = config
        self.metrics = metrics

    def publish_changes(self, event_msg):
        # do some publishing
        await self.metrics.incr('records-published', len(records))
```

Since [`incr`](https://github.com/spotify/aioshumway#initialize-a-counter-with-a-value) behavior in aioshumway does not yet emit the metrics, the interface implementation in `metrics.py` will have to overwrite the `incr` method to call `emit` right afterwords. This is future-proofing a little bit, allowing future metrics implementations that are pull-based (which prometheus is by default) to work, while still automatically emitting our own metrics. Sorry if that's confusing... 

**Release note**:

```release-note
Add interface definitions for a metrics plugin.
```

@spotify/alf

### DIFF
--- a/gordon/interfaces.py
+++ b/gordon/interfaces.py
@@ -120,3 +120,46 @@ class IPublisherClient(IGenericPlugin):
         Args:
             event_msg (IEventMessage): Message ready to be sent to destination.
         """
+
+
+class IMetricRelay(Interface):
+    """Create and publish (push) metrics on demand."""
+
+    def incr(metric_name, value=1):
+        """Increment a metric. Create one if new."""
+
+    def timer(metric_name):
+        """Return a :class:`ITimerMetric`. Create one if new."""
+
+    def set_counter(metric_name, counter):
+        """Set a :class:`ICounterMetric` object to a given metric name."""
+
+    def set_timer(metric_name, timer):
+        """Set a :class:`ITimerMetric` to a given metric name."""
+
+
+class IGenericMetric(Interface):
+    """**Do not** implement this interface directly.
+
+    Use :class:`ICounterMetric` or :class:`ITimerMetric`
+    instead.
+    """
+
+
+class ICounterMetric(IGenericMetric):
+    """Monotonically increasing counter object."""
+
+    value = Attribute('Count value, must start at 0.')
+
+    def incr(value=1):
+        """Increase counter by 1 or a given amount."""
+
+
+class ITimerMetric(IGenericMetric):
+    """Timer for a block of code."""
+    value = Attribute('Timed value, must be a positive number.')
+
+
+class IGaugeMetric(IGenericMetric):
+    """An instantaneous measurement of a value."""
+    value = Attribute('Measured value.')
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -38,3 +38,18 @@ interfaces
 
 .. autointerface:: gordon.interfaces.IGenericPlugin
     :members:
+
+.. autointerface:: gordon.interfaces.IMetricRelay
+    :members:
+
+.. autointerface:: gordon.interfaces.ICounterMetric
+    :members:
+
+.. autointerface:: gordon.interfaces.ITimerMetric
+    :members:
+
+.. autointerface:: gordon.interfaces.IGaugeMetric
+    :members:
+
+.. autointerface:: gordon.interfaces.IGenericMetric
+    :members:


econchick:
In sketching this out yesterday, it helped me to draft out the potential usage of the API defined in the interface. I made [this WIP branch](https://github.com/spotify/gordon/tree/wip/metrics-provider) - specifically a [metrics.py](https://github.com/spotify/gordon/blob/wip/metrics-provider/gordon/metrics.py) module - that shows my idea process, and could be used by whomever ends up implementing the interface for [aioshumway](https://github.com/spotify/aioshumway).

gordon/interfaces.py
```diff
@@ -120,3 +120,41 @@ class IPublisherClient(IGenericPlugin):
...
+
+    Use :class:`ICounterMetric` or :class:`ITimerMetric`
+    instead.
+    """
+
+
+class ICounterMetric(IGenericMetric):
+    """Monotonically increasing counter object."""
+
+    value = Attribute('Value, must start at 0.')
```
matt0bi: :thinking: What do you think about changing this attribute from `value` to `counter`? Or adding that word to the attribute's description, eg. _Counter. Must start at 0._?

`value` is nice in that all the metric objects (only Counter and Timer now) have a uniform interface. But `counter` and `timing` would be more descriptive at the cost of less uniformity. Thoughts?

gordon/interfaces.py
```diff
@@ -120,3 +120,41 @@ class IPublisherClient(IGenericPlugin):
...
+
+class IGenericMetric(Interface):
+    """**Do not** implement this interface directly.
+
+    Use :class:`ICounterMetric` or :class:`ITimerMetric`
+    instead.
+    """
+
+
+class ICounterMetric(IGenericMetric):
```
matt0bi: Do you think we will need a gauge-type metric as well for reporting changing numbers? For example, Gordon may jump between reporting a high number of messages (if the janitor sends out a lot of changes) and a low number if things are particularly quiet.

I'm not sure if a Counter metric would satisfy this situation in a long-running service like Gordon. I think that in cases like this we used to use shumway's Metric objects to fire off one off metrics.

gordon/interfaces.py
econchick: I'm cool for updating the description, but changing the actual attribute name to be more specific per metric would make it a bit more difficult to work with. It allows us to not know which object type we're working with but still call something like `update`, much like how shumway works (i.e. [counter](https://github.com/spotify/shumway/blob/master/shumway/__init__.py#L68), [timer](https://github.com/spotify/shumway/blob/master/shumway/__init__.py#L84)).

econchick:
@matt0bi updated! PTAL, thank you :)

gordon/interfaces.py
```diff
@@ -120,3 +120,46 @@ class IPublisherClient(IGenericPlugin):
...
+    """Monotonically increasing counter object."""
+
+    value = Attribute('Count value, must start at 0.')
+
+    def incr(value=1):
+        """Increase counter by 1 or a given amount."""
+
+
+class ITimerMetric(IGenericMetric):
```
matt0bi: One last Q: should the ITimerMetric iface specify a context manager interface? Or are we ok with timers sometimes _not_ being context managers?

econchick: I'm okay with them not being context managers, because they could be decorators ([prometheus example](https://github.com/prometheus/client_python#summary)).

matt0bi:
LGTM :+1:

gordon/interfaces.py
```diff
@@ -120,3 +120,46 @@ class IPublisherClient(IGenericPlugin):
         Args:
             event_msg (IEventMessage): Message ready to be sent to destination.
         """
+
+
+class IMetricRelay(Interface):
```
descent3me: What do you see the role of the metrics relay being in Gordon?  I've found limited use for it from Shumway and I'm concerned about codifying the design elsewhere. My thinking is that for our use case it makes sense to have an object that stores the socket information to ffwd, and an `emit` type method. For other users I can see the value of a some sort of internal data store of metrics and a `pull` method. I do also have a concern about trying to force two different use cases (socket store, emit) and (metrics store, pull) into the same abstraction, making it harder to understand both, since there seems to be very little overlap.

gordon/interfaces.py
```diff
@@ -120,3 +120,46 @@ class IPublisherClient(IGenericPlugin):
...
+    """Create and publish (push) metrics on demand."""
+
+    def incr(metric_name, value=1):
```
descent3me: For this and `timer`, I think having one method that does two things is of limited value and confusing. I also am worried that it is another layer of abstraction that has to be kept in step between the end user and the init method of the metric itself, which has shown to be a problem in Shumway.

gordon/interfaces.py
```diff
@@ -120,3 +120,46 @@ class IPublisherClient(IGenericPlugin):
...
+        """Increment a metric. Create one if new."""
+
+    def timer(metric_name):
+        """Return a :class:`ITimerMetric`. Create one if new."""
+
+    def set_counter(metric_name, counter):
```
descent3me: I don't understand why this and `set_timer` are separate methods, couldn't `add_metric(metric)` be just as useful? I also disagree that the metric name should be stored with the relay rather than the metric itself

gordon/interfaces.py
descent3me: This is outdated convo, right? There is a Gauge below.

PR #21, opened 2018-06-05
Update plugin_loader to load metrics plugin; Add FFWD metrics plugin

matt0bi:

This PR updates the plugin loader to pick and load a metrics and to pass it to all the other plugins. It also implements a simple FFWD-compatible metrics plugin that adheres to the IMetricRelay interface and which can be used to push metrics to FFWD. 

N/A

The FFWD plugin is included by default in order to avoid creating another repository. Since FFWD is our current choice for metrics, it makes sense if we include and support it as part of Gordon. Because users select the metrics plugin through the configuration file, including it doesn't mean users are forced to use FFWD.

The FFWD plugin doesn't use or depend on [aioshumway](https://github.com/spotify/aioshumway/). We opted for this in order to break away from the shumway interface, which we wouldn't use in any way apart from the `MetricRelay.emit` method, and to have one less dependency, which also affords us more flexibility. 

This PR doesn't include the default metrics plugin, which, instead of emitting metrics, will log them to stdout. The next PR will. 

**Release note**:

```release-note
Enable plugin loader to load metrics plugin.
Add FFWD-compatible metrics plugin.
```

@spotify/alf

### DIFF
--- a/gordon/interfaces.py
+++ b/gordon/interfaces.py
@@ -147,10 +147,8 @@ class ITimer(Interface):
     async def __aexit__(type, value, traceback):
         """Exit context manager to stop timing."""
 
-
     async def start():
         """Start timing."""
-
 
     async def stop():
         """End timing."""
--- a/gordon/interfaces.py
+++ b/gordon/interfaces.py
@@ -125,14 +125,32 @@ class IPublisherClient(IGenericPlugin):
 class IMetricRelay(Interface):
     """Manage Gordon metrics."""
 
-    def incr(metric_name, value=1):
+    async def incr(metric_name, value=1, **kwargs):
         """Increase counter metric by 1 or a given amount."""
 
-    def timer(metric_name):
-        """Time a block of code."""
+    def timer(metric_name, **kwargs):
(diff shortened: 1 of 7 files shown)


gordon/interfaces.py
```diff
@@ -125,14 +125,32 @@ class IPublisherClient(IGenericPlugin):
...
+    async def set(metric_name, value, **kwargs):
         """Set a gauge metric to a given value."""
 
-    def cleanup():
+    async def cleanup(**kwargs):
         """Perform cleanup tasks related to metrics handling."""
+
+
+class ITimer(Interface):
+    """Timer support]ing both manual operation and use as a context manager."""
```
econchick: Typo (`]`)

matt0bi:
Closing in order to to make a new PR (#22) that adheres to our branch naming flow.

descent3me:
D'oh!  Good thinking about the branch naming @matt0bi! I forgot about that.

matt0bi:
@descent3me it was @econchick's eagle eyes that caught this :)

PR #22, opened 2018-06-05
Update plugin_loader to load metrics plugin; Add FFWD metrics plugin

matt0bi:

This PR updates the plugin loader to pick and load a metrics and to pass it to all the other plugins. It also implements a simple FFWD-compatible metrics plugin that adheres to the IMetricRelay interface and which can be used to push metrics to FFWD. 

Note, this is the same PR as #21 but with the branch changed from `add_metrics` to `feature/add_metrics` to comply with the contributing guide.

N/A

The FFWD plugin is included by default in order to avoid creating another repository. Since FFWD is our current choice for metrics, it makes sense if we include and support it as part of Gordon. Because users select the metrics plugin through the configuration file, including it doesn't mean users are forced to use FFWD.

The FFWD plugin doesn't use or depend on [aioshumway](https://github.com/spotify/aioshumway/). We opted for this in order to break away from the shumway interface, which we wouldn't use in any way apart from the `MetricRelay.emit` method, and to have one less dependency, which also affords us more flexibility. 

This PR doesn't include the default metrics plugin, which, instead of emitting metrics, will log them to stdout. The next PR will. 

**Release note**:

```release-note
Enable plugin loader to load metrics plugin.
Add FFWD-compatible metrics plugin.
```

@spotify/alf

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -48,6 +48,11 @@ core
 
     If ``false``, Gordon will exit out if it can't load one or more plugins.
 
+.. option:: metrics="STR"
+
+    The metrics provider to use. Depending on the provider, more
+    configuration may be needed. See provider implementation for details.
+
 
 core.logging
 ~~~~~~~~~~~~
--- a/gordon.toml.example
+++ b/gordon.toml.example
@@ -2,6 +2,7 @@
 [core]
 plugins = ["foo.plugin"]
 debug = false
+metrics = "ffwd"
 
 [core.logging]
 level = "info"
--- a/gordon/interfaces.py
+++ b/gordon/interfaces.py
@@ -61,7 +61,7 @@ class IGenericPlugin(Interface):
     """
     phase = Attribute('Plugin phase')
 
-    def __init__(config, success_channel, error_channel, metrics=None):
+    def __init__(config, success_channel, error_channel, metrics):
         """Initialize an EventClient object.
 
         Args:
--- a/gordon/interfaces.py
+++ b/gordon/interfaces.py
@@ -125,14 +125,50 @@ class IPublisherClient(IGenericPlugin):
 class IMetricRelay(Interface):
     """Manage Gordon metrics."""
 
-    def incr(metric_name, value=1):
-        """Increase counter metric by 1 or a given amount."""
+    async def incr(metric_name, value=1, context=None, **kwargs):
+        """Increase a metric by 1 or a given amount.
 
-    def timer(metric_name):
-        """Time a block of code."""
+        Args:
+            metric_name (str): Identifier of the metric.
+            value (int): (optional) Value with which to increase the metric.
+            context (dict): (optional) Additional key-value pairs which further
+                describe the metric, for example: {'remote-host': '1.2.3.4'}
+        """
 
-    def set(metric_name, value):
-        """Set a gauge metric to a given value."""
+    def timer(metric_name, context=None, **kwargs):
+        """Get a timer object which implements ITimer.
 
-    def cleanup():
+        Args:
+            metric_name (str): Identifier of the metric.
+            context (dict): (optional) Additional key-value pairs which further
+                describe the metric, for example: {'unit': 'seconds'}
+        """
+
+    async def set(metric_name, value, context=None, **kwargs):
+        """Set a metric to a given value.
+
+        Args:
+            metric_name (str): Identifier of the metric.
+            value (number): The value of the metric.
+            context (dict): (optional) Additional key-value pairs which further
+                describe the metric, for example: {'app-version': '1.5.3'}
+        """
+
+    async def cleanup(**kwargs):
         """Perform cleanup tasks related to metrics handling."""
+
+
+class ITimer(Interface):
+    """Timer supporting both manual operation and use as a context manager."""
+
+    async def __aenter__():
+        """Enter context manager to start timing."""
+
+    async def __aexit__(type, value, traceback):
+        """Exit context manager to stop timing."""
+
+    async def start():
+        """Start timing."""
+
+    async def stop():
+        """End timing."""
--- a/gordon/metrics/__init__.py
+++ b/gordon/metrics/__init__.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2018 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
--- a/gordon/metrics/ffwd.py
+++ b/gordon/metrics/ffwd.py
@@ -1,0 +1,201 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+
+Gordon ships with a simple `ffwd <https://github.com/spotify/ffwd>`_
+metrics implementation, which can be enabled via configuration. This
+module contains the SimpleFfwdRelay, and all required classes that it
+uses to send messsages to the ffwd daemon via UDP.
+
+The `SimpleFfwdRelay` requires no configuration, but can be customized.
+The defaults that may be overridden are shown below.
+
+.. code-block:: ini
+
+    [ffwd]
+    # to identify the service creating metrics
+    key = 'gordon-service'
+
+    # the address of the ffwd daemon (see: UDPClient)
+    ip = "127.0.0.9"
+    port = 19000
+
+    # a scaling factor for timing (see: FfwdTimer)
+    time_unit = 1E9
+
+"""
+
+import asyncio
+import json
+import time
+
+import zope.interface
+
+from gordon import interfaces
+
+
+class UDPClientProtocol(asyncio.DatagramProtocol):
+    """Protocol for sending one-off messages via UDP.
+
+    Args:
+        message (bytes): Message for ffwd agent.
+    """
+    def __init__(self, message):
+        self.message = message
+        self.transport = None
+
+    def connection_made(self, transport):
+        """Create connection, use to send message and close.
+
+        Args:
+            transport (asyncio.DatagramTransport): Transport used for sending.
+        """
+        self.transport = transport
+        self.transport.sendto(self.message)
+        self.transport.close()
+
+
+class UDPClient:
+    """Client for sending UDP datagrams.
+
+    Args:
+        ip (str): (optional) Destination IP address (default: 127.0.0.1).
+        port (int): (optional) Destination port (default: 9000).
+        loop (asyncio.AbstractEventLoop impl): (optional) Event loop.
+    """
+
+    DEFAULT_IP = '127.0.0.1'
+    DEFAULT_PORT = 19000
+
+    def __init__(self, ip=None, port=None, loop=None):
+        self.ip = ip or self.DEFAULT_IP
+        self.port = port or self.DEFAULT_PORT
+        self.loop = loop or asyncio.get_event_loop()
+
+    async def send(self, metric):
+        """Transform metric to JSON bytestring and send to server.
+
+        Args:
+            metric (dict): Complete metric to send as JSON.
+        """
+        message = json.dumps(metric).encode('utf-8')
+        await self.loop.create_datagram_endpoint(
+            lambda: UDPClientProtocol(message),
+            remote_addr=(self.ip, self.port))
+
+
+@zope.interface.implementer(interfaces.ITimer)
+class FfwdTimer:
+    """Timer which sends UDP messages to FFWD on completion.
+
+    Args:
+        metric (dict): Dict representation of the metric to send.
+        udp_client (UDPClient): A metric sending client.
+        time_unit (number): (optional) Scale time unit for use with
+            time.perf_counter(), for example: 1E9 to send nanoseconds.
+    """
+    def __init__(self, metric, udp_client, time_unit=None):
+        self.metric = metric
+        self.udp_client = udp_client
+        self.time_unit = time_unit or 1
+
+    async def __aenter__(self):
+        """Enter context manager to start timing."""
+        await self.start()
+        return self
+
+    async def __aexit__(self, *args):
+        """Exit context manager to stop timing."""
+        await self.stop()
+
+    async def start(self):
+        """Start timer."""
+        self._start_time = time.perf_counter()
+
+    async def stop(self):
+        """Stop timer."""
+        time_elapsed = time.perf_counter() - self._start_time
+        self.metric['value'] = time_elapsed * self.time_unit
+        await self.udp_client.send(self.metric)
+
+
+@zope.interface.implementer(interfaces.IMetricRelay)
+class SimpleFfwdRelay:
+    """Metrics relay which sends to FFWD immediately.
+
+    The relay does no client-side aggregation and metrics are
+    emitted immediately. The relay uses a combination of the `key and
+    attributes fields <https://github.com/spotify/ffwd/tree/master/
+    modules/json>`_ to semantically identify metrics in ffwd.
+
+    Args:
+        config (dict): Configuration with optional keys described above.
+    """
+    def __init__(self, config):
+        self.key = config.get('key', 'gordon-service')
+        self.udp_client = UDPClient(
+            config.get('ffwd_ip'), config.get('ffwd_port'))
+        self.time_unit = config.get('time_unit')
+
+    def _create_metric(self, metric_name, value, context, **kwargs):
+        attrs = context or {}
+        attrs['what'] = metric_name
+
+        metric = {
+            'key': self.key,
+            'attributes': attrs,
+            'value': value,
+            'type': 'metric'
+        }
+        return metric
+
+    async def incr(self, metric_name, value=1, context=None, **kwargs):
+        """Increase a metric by 1 or a given amount.
+
+        Args:
+            metric_name (str): Identifier of the metric.
+            value (int): (optional) Value with which to increase the metric
+                (default: 1).
+            context (dict): (optional) Additional key-value pairs which further
+                describe the metric, for example: {'remote-host': '1.2.3.4'}
+        """
+        await self.set(metric_name, value, context, **kwargs)
+
+    def timer(self, metric_name, context=None, **kwargs):
+        """Create a FfwdTimer.
+
+        Args:
+            metric_name (str): Identifier of the metric.
+            context (dict): (optional) Additional key-value pairs which further
+                describe the metric, for example: {'unit': 'seconds'}
+        """
+        metric = self._create_metric(metric_name, None, context, **kwargs)
+        return FfwdTimer(metric, self.udp_client, self.time_unit)
+
+    async def set(self, metric_name, value, context=None, **kwargs):
+        """Set a metric to a given value.
+
+        Args:
+            metric_name (str): Identifier of the metric.
+            value (number): The value of the metric.
+            context (dict): (optional) Additional key-value pairs which further
+                describe the metric, for example: {'app-version': '1.5.3'}
+        """
+        metric = self._create_metric(metric_name, value, context, **kwargs)
+        await self.udp_client.send(metric)
+
+    async def cleanup(self):
+        """Not used."""
+        pass
--- a/gordon/plugins_loader.py
+++ b/gordon/plugins_loader.py
@@ -59,6 +59,7 @@ import logging
 import pkg_resources
 
 from gordon import exceptions
+from gordon.metrics import ffwd
 
 
 PLUGIN_NAMESPACE = 'gordon.plugins'
@@ -191,7 +192,9 @@ def _get_metrics_plugin(config, installed_plugins):
     if not metrics_provider:
         return
 
-    metrics_config = config.get(metrics_provider)
+    metrics_config = config.get(metrics_provider, {})
+    if metrics_provider == 'ffwd':
+        return ffwd.SimpleFfwdRelay(metrics_config)
 
     for plugin_name, plugin in installed_plugins.items():
         if metrics_provider == plugin_name:
--- a/gordon/plugins_loader.py
+++ b/gordon/plugins_loader.py
@@ -186,6 +186,22 @@ def _gather_installed_plugins():
     return gathered_plugins
 
 
+def _get_metrics_plugin(config, installed_plugins):
+    metrics_provider = config.get('core', {}).get('metrics')
+    if not metrics_provider:
+        return
+
+    metrics_config = config.get(metrics_provider)
+
+    for plugin_name, plugin in installed_plugins.items():
+        if metrics_provider == plugin_name:
+            plugin_class = plugin.load()
+            return plugin_class(metrics_config)
+
+    msg = f'Metrics Plugin "{metrics_provider}" configured, but not installed'
+    raise exceptions.LoadPluginError(msg)
+
+
 def load_plugins(config, plugin_kwargs):
     """
     Discover and instantiate plugins.
@@ -202,6 +218,10 @@ def load_plugins(config, plugin_kwargs):
         config.
     """
     installed_plugins = _gather_installed_plugins()
+    metrics_plugin = _get_metrics_plugin(config, installed_plugins)
+    if metrics_plugin:
+        plugin_kwargs['metrics'] = metrics_plugin
+
     active_plugins = _get_activated_plugins(config, installed_plugins)
     if not active_plugins:
         return [], [], []
--- a/gordon/plugins_loader.py
+++ b/gordon/plugins_loader.py
@@ -227,8 +227,9 @@ def load_plugins(config, plugin_kwargs):
 
     active_plugins = _get_activated_plugins(config, installed_plugins)
     if not active_plugins:
-        return [], [], []
+        return [], [], [], None
     plugin_namespaces = _get_plugin_config_keys(active_plugins)
     plugin_configs = _load_plugin_configs(plugin_namespaces, config)
-    return _init_plugins(
+    plugin_names, plugins, errors = _init_plugins(
         active_plugins, installed_plugins, plugin_configs, plugin_kwargs)
+    return plugin_names, plugins, errors, metrics_plugin
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -41,7 +41,6 @@ class EventConsumerStub:
         self.config = config
         self.success_channel = success_channel
         self.error_channel = error_channel
-        self.metrics = metrics
         self._mock_run_count = 0
         self._mock_cleanup_count = 0
 
@@ -64,7 +63,6 @@ class EnricherStub:
         self.success_channel = success_channel
         self.error_channel = error_channel
         self._mock_process_count = 0
-        self.metrics = metrics
 
     async def process(self, event_msg):
         await asyncio.sleep(0)
@@ -81,7 +79,6 @@ class PublisherStub:
         self.success_channel = success_channel
         self.error_channel = error_channel
         self._mock_publish_changes_count = 0
-        self.metrics = metrics
 
     async def publish_changes(self, event_msg):
         await asyncio.sleep(0)
@@ -96,7 +93,6 @@ class GenericStub:
         self.config = config
         self.success_channel = success_channel
         self.error_channel = error_channel
-        self.metrics = metrics
         self._mock_run_count = 0
 
     async def run(self):
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -37,10 +37,11 @@ REGISTERED_ACTIVE_PLUGINS = REGISTERED_PLUGINS[:3]
 
 @zope.interface.implementer(interfaces.IEventConsumerClient)
 class EventConsumerStub:
-    def __init__(self, config, success_channel, error_channel, **kwargs):
+    def __init__(self, config, success_channel, error_channel, metrics=None):
         self.config = config
         self.success_channel = success_channel
         self.error_channel = error_channel
+        self.metrics = metrics
         self._mock_run_count = 0
         self._mock_cleanup_count = 0
 
@@ -58,11 +59,12 @@ class EventConsumerStub:
 
 @zope.interface.implementer(interfaces.IEnricherClient)
 class EnricherStub:
-    def __init__(self, config, success_channel, error_channel, **kwargs):
+    def __init__(self, config, success_channel, error_channel, metrics=None):
         self.config = config
         self.success_channel = success_channel
         self.error_channel = error_channel
         self._mock_process_count = 0
+        self.metrics = metrics
 
     async def process(self, event_msg):
         await asyncio.sleep(0)
@@ -74,11 +76,12 @@ class EnricherStub:
 
 @zope.interface.implementer(interfaces.IPublisherClient)
 class PublisherStub:
-    def __init__(self, config, success_channel, error_channel, **kwargs):
+    def __init__(self, config, success_channel, error_channel, metrics=None):
         self.config = config
         self.success_channel = success_channel
         self.error_channel = error_channel
         self._mock_publish_changes_count = 0
+        self.metrics = metrics
 
     async def publish_changes(self, event_msg):
         await asyncio.sleep(0)
@@ -89,10 +92,11 @@ class PublisherStub:
 
 
 class GenericStub:
-    def __init__(self, config, success_channel, error_channel, **kwargs):
+    def __init__(self, config, success_channel, error_channel, metrics=None):
         self.config = config
         self.success_channel = success_channel
         self.error_channel = error_channel
+        self.metrics = metrics
         self._mock_run_count = 0
 
     async def run(self):
--- a/tests/unit/metrics/test_ffwd.py
+++ b/tests/unit/metrics/test_ffwd.py
@@ -1,0 +1,203 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numbers
+
+import pytest
+
+from gordon.metrics import ffwd
+
+
+def test_ffwd_protocol_connection_made(mocker):
+    """Message sends and connection closes."""
+    mock_transport = mocker.Mock()
+    message = 'Testing message!'
+
+    protocol = ffwd.UDPClientProtocol(message)
+    protocol.connection_made(mock_transport)
+
+    mock_transport.sendto.assert_called_once_with(message)
+    mock_transport.close.assert_called_once_with()
+
+
+@pytest.mark.asyncio
+async def test_udp_client_send(mocker):
+    """Sends serialized metric using FfwdClientProtocol."""
+    mock_loop = mocker.Mock()
+
+    actual_func = None
+    actual_addr = None
+
+    async def mock_create_datagram_endpoint(func, remote_addr):
+        nonlocal actual_func
+        nonlocal actual_addr
+        actual_func = func
+        actual_addr = remote_addr
+
+    mock_loop.create_datagram_endpoint = mock_create_datagram_endpoint
+
+    udp_client = ffwd.UDPClient('1.2.3.4', '5678', mock_loop)
+    await udp_client.send({'an': 'object'})
+
+    actual_protocol = actual_func()
+    assert isinstance(actual_protocol, ffwd.UDPClientProtocol)
+    assert ('1.2.3.4', '5678') == actual_addr
+    assert b'{"an": "object"}' == actual_protocol.message
+
+
+@pytest.fixture
+def mock_udp_client_and_metrics_sent(mocker):
+    metrics_sent = []
+
+    async def mock_send(metric):
+        nonlocal metrics_sent
+        metrics_sent.append(metric)
+
+    mock_udp_client = mocker.Mock()
+    mock_udp_client.send = mock_send
+    return mock_udp_client, metrics_sent
+
+
+@pytest.fixture
+def timer_and_metrics_sent(mock_udp_client_and_metrics_sent):
+    mock_udp_client, metrics_sent = mock_udp_client_and_metrics_sent
+    metric = {'a': 'metric', 'value': None}
+    timer = ffwd.FfwdTimer(metric, mock_udp_client)
+    return timer, metrics_sent
+
+
+@pytest.mark.asyncio
+async def test_ffwdtimer_manual(timer_and_metrics_sent):
+    """Times when operated manually."""
+    timer, metrics_sent = timer_and_metrics_sent
+    await timer.start()
+    assert isinstance(timer._start_time, numbers.Number)
+
+    await timer.stop()
+    assert 1 == len(metrics_sent)
+    assert 'metric' == metrics_sent[0]['a']
+    assert 0 < metrics_sent[0]['value']
+
+
+@pytest.mark.asyncio
+async def test_ffwdtimer_as_context_manager(timer_and_metrics_sent):
+    """Times as a context manager."""
+    timer, metrics_sent = timer_and_metrics_sent
+    async with timer as t:
+        assert isinstance(t._start_time, numbers.Number)
+
+    assert 1 == len(metrics_sent)
+    assert 'metric' == metrics_sent[0]['a']
+    assert 0 < metrics_sent[0]['value']
+
+
+@pytest.fixture
+def test_config():
+    return {
+        'key': 'service-name',
+        'ffwd_ip': '1.2.3.4',
+        'ffwd_port': '5678',
+        'time_unit': 1E9
+    }
+
+
+TEST_KEY_NAME_VAL = [
+    ('service-name', 'metric-name', 10),
+    ('other-name', 'metric-name', 6.243289),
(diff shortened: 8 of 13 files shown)


tests/unit/metrics/test_ffwd.py
```diff
@@ -0,0 +1,208 @@
...
+async def test_ffwdtimer_manual(timer_and_metrics_sent):
+    """Times when operated manually."""
+    timer, metrics_sent = timer_and_metrics_sent
+    await timer.start()
+    assert isinstance(timer._start_time, numbers.Number)
+
+    await timer.stop()
+    assert len(metrics_sent) == 1
+    assert 'metric' == metrics_sent[0]['a']
+    assert 0 < metrics_sent[0]['value']
```
ocsw: I feel like this and the same line below would be a lot easier to read as foo > 0.  But if you're going to do it this way, the length comparison should be in the other order.

tests/unit/metrics/test_ffwd.py
```diff
@@ -0,0 +1,208 @@
...
+    async with timer as t:
+        assert isinstance(t._start_time, numbers.Number)
+
+    assert len(metrics_sent) == 1
+    assert 'metric' == metrics_sent[0]['a']
+    assert 0 < metrics_sent[0]['value']
+
+
+@pytest.fixture
+def test_config(mocker, monkeypatch):
```
ocsw: The name should indicate that it also has a mock.  Or do the mock separately.

tests/unit/metrics/test_ffwd.py
```diff
@@ -0,0 +1,208 @@
...
+@pytest.fixture
+def ffwdrelay_and_metrics_sent(test_config, mock_udp_client_and_metrics_sent):
+    mock_udp_client, metrics_sent = mock_udp_client_and_metrics_sent
+    relay = ffwd.FfwdRelay(test_config)
+    relay.udp_client = mock_udp_client
+    return relay, metrics_sent
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize('attrs,tags', [
```
ocsw: Maybe pull this array out instead of repeating it?

tests/unit/metrics/test_ffwd.py
```diff
@@ -0,0 +1,208 @@
...
+    (None, None),
+    ({'hostname': 'dns-host-3'}, None),
+    (None, ['v40']),
+    ({'type': 'active'}, ['v40'])
+])
+async def test_ffwdrelay_set(ffwdrelay_and_metrics_sent, attrs, tags):
+    """Sends metric."""
```
ocsw: I would make this 'Initializes and sends', like the incr test.

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,138 @@
...
+        self.message = message
+        self.transport = None
+
+    def connection_made(self, transport):
+        self.transport = transport
+        self.transport.sendto(self.message)
+        self.transport.close()
+
+
+class UDPClient:
```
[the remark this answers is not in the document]
econchick: [It's actually dropped in py3](http://portingguide.readthedocs.io/en/latest/classes.html)

gordon/interfaces.py
```diff
@@ -125,14 +125,30 @@ class IPublisherClient(IGenericPlugin):
...
+    async def set(metric_name, value, **kwargs):
         """Set a gauge metric to a given value."""
 
-    def cleanup():
+    async def cleanup(**kwargs):
         """Perform cleanup tasks related to metrics handling."""
+
+
+class ITimer(Interface):
+    """Timer support]ing both manual operation and use as a context manager."""
```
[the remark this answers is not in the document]
descent3me: Dang it! Will fix!

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,138 @@
...
+        self.metric['value'] = time_elapsed
+        await self.udp_client.send(self.metric)
+
+
+@zope.interface.implementer(interfaces.IMetricRelay)
+class FfwdRelay:
+    """IMetricRelay implementation which sends metrics to FFWD immediately.
+
+    Args:
+        config (dict): Required keys:
```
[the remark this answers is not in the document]
descent3me: My intention was to wait until there's a standardized way to do it.

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,138 @@
...
+            ffwd_port (int): ffwd port number.
+            time_unit (number): Scale unit to use with timers.
+    """
+    def __init__(self, config):
+        self.key = config['key']
+        self.udp_client = UDPClient(config['ffwd_ip'], config['ffwd_port'])
+        self.time_unit = config['time_unit']
+
+    def _create_metric(self, metric_name, value, **kwargs):
+        attrs = kwargs.get('attrs') or {}
```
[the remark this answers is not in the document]
descent3me: If someone passes is explicitly passes`attrs=None`, I want to be sure I don't put None in the json.  I actually don't know for sure that ffwd would freak out. I can look into it.

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,138 @@
...
+        return {
+            'key': self.key,
+            'attributes': attrs,
+            'value': value,
+            'type': 'metric',
+            'tags': tags
+        }
+
+    async def incr(self, metric_name, value=1, **kwargs):
+        await self.set(metric_name, value, **kwargs)
```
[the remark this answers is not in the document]
descent3me: Well, our increment is actually server-side, so we just emit `1` a bunch of times and do the sum in the Alien graph. Hmm maybe this should be called AlienFfwdRelay or at least include that in one of the docstrings?  @matt0bi Thoughts?

gordon/plugins_loader.py
```diff
@@ -186,6 +186,14 @@ def _gather_installed_plugins():
     return gathered_plugins
 
 
+def _get_metrics_plugin(config, installed_plugins):
+    metrics_config = config.get('metrics', {})
+    for plugin_name, plugin in installed_plugins.items():
+        if plugin_name == metrics_config.get('provider', {}):
```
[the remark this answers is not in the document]
descent3me: Nope!  The only way we want that is if someone makes a ComboRelay that itself instantiates the two other plugins.

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
ocsw: Yeah, last I heard it was good practice to make it explicit, but I guess that doesn't matter if we require 3.6.

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
ocsw: Eh, fair enough.

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
ocsw: ok

gordon/plugins_loader.py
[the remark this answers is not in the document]
ocsw: ok

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
econchick: What clarity does it bring for you?

Might have been good practice for py2, but never for py3. It's one of the reasons folks use py3.

tests/unit/test_plugins_loader.py
```diff
@@ -272,3 +274,66 @@ def test_load_plugins_exceptions(installed_plugins, loaded_config,
...
+def metrics_mock(mocker):
+    relay_mock = mocker.MagicMock(pkg_resources.EntryPoint)
+    relay_mock.name = 'gordon.plugins.metrics.mock'
+    relay_mock.load.return_value = MetricRelayStub
+    return relay_mock
+
+
+@pytest.fixture
+def plugins_incl_metrics(mocker, monkeypatch, metrics_mock, installed_plugins):
+    all_plugins = installed_plugins.copy()
```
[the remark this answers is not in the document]
econchick: Agreed w/ this comment. 

But just FTR so there's not confusion: [fixtures are scoped](https://docs.pytest.org/en/latest/fixture.html#scope-sharing-a-fixture-instance-across-tests-in-a-class-module-or-session), by `function` by default (aka the fixture function is called every time it's used). You can scope to `class`, `module` (only call the function once per class/module and use the same result within the class/module), and `session` (once per test run).

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
matt0bi: Yeah, since py3 classes inherit from `object` by default and we don't care for backwards compatibility, omitting `(object)` seems more clear to me.

gordon/plugins_loader.py
econchick: Hm, this forces the service operator to be aware of the entry point name that the plugin itself has declared.

Could we have the config key/section be the name without the `gordon.metrics` prefix, i.g. `[ffwd]`? Then maybe add another optional `[core]` variable, e.g.

```toml
[core]
plugins = ["gcp.publisher", "..."]
# or metrics_provider, or whatever
metrics = "ffwd"

[ffwd]
host = "10.99.0.1"
port = 19000
```

so then the code above is similar:

```py
metrics_provider = config.get('metrics')

# maybe this part is at a later commit/PR?
if not metrics_provider:  
    # instantiate & return metrics logger
if metrics_provider == 'ffwd'
    # import ffwd, instantiate w/ config, & return

for plugin_name, plugin in installed_plugins.items():
   if plugin_name == metrics_provider:
        metrics_config = config.get(metrics_provider, {})
        # load, instantiate, return
```

This allows configuration to be abstracted away a bit better:

```toml
[core]
metrics = "prometheus"

[prometheus]
exporter = "http"  # or twisted, wsgi, pushgateway
multiprocess = true
authentication = "yesplease"
```

And plugin developers can still register their entry point w/ a simpler name:

```py
# prometheus_plugin/setup.py
setup(
    # snip
    entry_points={
        'gordon.plugins': [
            'prometheus = path.to:callable',
            # would otherwise be this w/ current setup
            'gordon.metrics.prometheus = path.to.callable'
        ],
    },
    # snip
)
```

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
matt0bi: >  Hmm maybe this should be called AlienFfwdRelay or at least include that in one of the docstrings?

To clarify, FFWD + Heroic can be used with front-ends different than Alien, but in this case we're completely relying on Alien to do the accumulation for us? I know at least one other tsdb that uses the same approach (influxdb and its derivative fn), so I think including something in the docstring about this should be fine.

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        message = json.dumps(metric).encode('utf-8')
+        await self.loop.create_datagram_endpoint(
+            lambda: FfwdClientProtocol(message),
+            remote_addr=(self.ip, self.port))
+
+
+@zope.interface.implementer(interfaces.ITimer)
+class FfwdTimer:
+    """ITimer implementation which sends UDP messages to FFWD on completion.
```
econchick: This (and other implementer) docstrings could be improved; we already know it implemented these interfaces. Maybe something like """Timer metric emitted immediately upon completion."""

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,138 @@
...
+@zope.interface.implementer(interfaces.IMetricRelay)
+class FfwdRelay:
+    """IMetricRelay implementation which sends metrics to FFWD immediately.
+
+    Args:
+        config (dict): Required keys:
+            key (str): A key to identify the service as a whole.
+            ffwd_ip (str): ffwd IP address.
+            ffwd_port (int): ffwd port number.
+            time_unit (number): Scale unit to use with timers.
```
econchick: Maybe a default can be set?

tests/unit/metrics/test_ffwd.py
```diff
@@ -0,0 +1,208 @@
...
+
+@pytest.mark.asyncio
+async def test_ffwdtimer_manual(timer_and_metrics_sent):
+    """Times when operated manually."""
+    timer, metrics_sent = timer_and_metrics_sent
+    await timer.start()
+    assert isinstance(timer._start_time, numbers.Number)
+
+    await timer.stop()
+    assert len(metrics_sent) == 1
```
econchick: Don't forget: `assert expected == actual`

gordon/plugins_loader.py
```diff
@@ -202,6 +210,10 @@ def load_plugins(config, plugin_kwargs):
         config.
     """
     installed_plugins = _gather_installed_plugins()
+    metrics_plugin = _get_metrics_plugin(config, installed_plugins)
+    if metrics_plugin:
+        plugin_kwargs['metrics'] = metrics_plugin
+
```
econchick: Is there a need for an `if metrics_plugin` check? or rather, will this check be removed once the metric logger is implemented?

Could you also return the metrics plugin? The router will need it. It could just be a part of the list of inited plugins that are returned, and I can just pull it out by checking if it provides the interface.

tests/unit/metrics/test_ffwd.py
descent3me: I agree it would be easier to read (which is probably why I messed up the len() one, since I usually do it that way first out of habit), but our [contributing guidelines](https://github.com/spotify/gordon/blob/develop/CONTRIBUTING.rst#code) say we should do it the other way around. Will fix!

tests/unit/metrics/test_ffwd.py
descent3me: Hmm I don't even remember including the monkeypatch. I don't think it's used. I'll remove it.

tests/unit/metrics/test_ffwd.py
descent3me: Fair enough.

tests/unit/metrics/test_ffwd.py
descent3me: Actually, I might change the wording of both, since I'm not sure initialize is quite correct either.

tests/unit/metrics/test_ffwd.py
```diff
@@ -0,0 +1,208 @@
...
+        ffwdrelay._create_metric('some-counter', 42, attrs=attrs, tags=tags)
+    ]
+    await ffwdrelay.set('some-counter', 42, attrs=attrs, tags=tags)
+    assert expected == metrics_sent
+
+
+@pytest.mark.asyncio
+async def test_cleanup(ffwdrelay):
+    """Raises no errors."""
+    await ffwdrelay.cleanup()
```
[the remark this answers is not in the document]
descent3me: Technically the way it's written now, won't make tests fail since they'll just error out, but the build will go red, which I generally think is good enough.  @econchick any thoughts?

tests/unit/test_plugins_loader.py
```diff
@@ -272,3 +274,66 @@ def test_load_plugins_exceptions(installed_plugins, loaded_config,
...
+    all_plugins['gordon.plugins.metrics.mock'] = metrics_mock
+    mock_iter_entry_points = mocker.Mock(pkg_resources.iter_entry_points)
+    mock_iter_entry_points.return_value = iter(all_plugins.values())
+    monkeypatch.setattr(plugins_loader.pkg_resources, 'iter_entry_points',
+                        mock_iter_entry_points)
+    return all_plugins
+
+
+def test_load_plugins_with_metrics(plugins_incl_metrics, loaded_config,
```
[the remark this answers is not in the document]
descent3me: That happens up on https://github.com/spotify/gordon/pull/22/files#diff-2c18a76758cf57e7caa04cf54c62560eR238

tests/unit/test_plugins_loader.py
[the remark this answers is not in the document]
descent3me: True, we were seeing some weird behavior after we added this test, which is why we tried this. I think we found the cause elsewhere and just forgot to change it back. Will fix!

tests/unit/test_plugins_loader.py
```diff
@@ -272,3 +274,66 @@ def test_load_plugins_exceptions(installed_plugins, loaded_config,
...
+    config = {'metrics': {'provider': 'gordon.plugins.metrics.mock'}}
+    actual = plugins_loader._get_metrics_plugin(config, plugins_incl_metrics)
+    assert isinstance(actual, MetricRelayStub)
+
+
+@pytest.mark.parametrize('config,plugin_name', [
+    ({}, 'gordon.plugins.metrics.mock'),
+    ({'metrics': {'provider': 'gordon.plugins.metrics.mock'}}, 'other_name')
+])
+def test_get_metrics_fails(mocker, installed_plugins, metrics_mock, config,
```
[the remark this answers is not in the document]
descent3me: I think I might have a better one, but I'll rename!

gordon.toml.example
```diff
@@ -15,3 +15,10 @@ bar = baz
 ["foo.plugin"]
 # specific plugin config within "foo" package
 baz = bla
+
+[metrics]
```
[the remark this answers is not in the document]
descent3me: I agree, but since I think we're still not 100% sure this is the way we want to do metrics, I'd prefer to get the logic approved before writing a bunch of docs about it.

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
econchick: Hmm maybe I'm under-caffeinated, so please correct me if I'm wrong: as I'm instrumenting gordon core, I'm doing something like this:

```py
attrs = {'error': 'invalid-msg-provider'}
self.metrics.incr('dropped-message', attrs=attrs)
```

The `attrs` kwarg is specific to ffwd/our use case, so it seems like it's a bit leaky, i.e. if we switch for prometheus implementation, `attrs` will be ignored (hopefully not throw an error), and therefore losing valuable context.

I don't know how to handle tags (actually, I don't know if we really use them?), but perhaps we can assume all kwargs are attributes?

And maybe it'd be more clear if we renamed `kwargs` to `context` or `extra_context`?

tests/unit/metrics/test_ffwd.py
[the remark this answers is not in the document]
econchick: Yeah I'm cool with how it is now.

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
matt0bi: > if we switch for prometheus implementation, attrs will be ignored

I think prometheus wouldn't ignore attrs since it's similar to the concept of [labels][0] - key:values. We could rename it something more generic, but clearer (_attributes_?) to better fit how different dbs call it (heroic - attributes, prometheus - labels, influx - fields) and maybe make it a required argument?

[0]: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
ocsw: I guess I'm used to there always being parens, and looking at them to determine if the class has a superclass or not, vs this is with a thing or without.  I agree it's cleaner.  Guess I'll just have to adjust my habits.

gordon.toml.example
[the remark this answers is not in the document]
ocsw: Fair enough.  Let's just make sure there's a task for docs somewhere.

tests/unit/metrics/test_ffwd.py
ocsw: So just to clarify though...  Does 0 count as the 'expected' value if it's a comparison?

gordon/plugins_loader.py
ocsw: I'd like Gordon to be able to run with no metric plugin configured.  Unless we don't want to do that, in which case we should explicitly raise an error somewhere.

econchick: @ocsw - @matt0bi's PR (#23) is the default metrics provider, so it should always run with a metrics provider. @descent3me and I also chatted a bit and agreed that plugins can assume there will always be a metrics provider, so there's no `if self.metrics` littered throughout. The instrumentation I'm doing right now requires that core have a metrics provider too (since we want to provide some default metrics out of the box and not rely on plugin providers to do _everything).

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
descent3me: I was also thinking alien was too specific for a name, but I don't think specifying a bit more would be bad. How about something like `FastFfwdRelay` or some better name (`NonAggregatingFfwdRelay` seems too long)?  The thing I'm trying to clarify from the name is to differentiate the metrics are sent immediately, since it would be perfectly valid to set up a ffwd implementation that had a layer of aggregation in front of it (i.e. has an internal counter, instead of sending only 1's).

gordon/metrics/ffwd.py
descent3me: Ah true, I added a default to timer, but forgot to put it here.  Will fix!

gordon/plugins_loader.py
ocsw: I like this.  It's also more consistent with the config style for the other plugins.

gordon/plugins_loader.py
ocsw: ok

gordon/metrics/ffwd.py
descent3me: Yes, if you're looking at the source code it's obvious, but I was thinking for someone just using the library having the implementation note would be nice to have in the docstring, for example so it shows in `help(FfwdTimer)`, since I don't think the zope things show up there. But I can remove it if you think its excessive.

tests/unit/test_plugins_loader.py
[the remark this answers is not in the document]
ocsw: This link seems to have broken...

gordon/metrics/ffwd.py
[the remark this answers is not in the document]
descent3me: Yeah, I was going for non-breaking, but that data loss was acceptable.  I wasn't sure if all other libraries supported additional data, so I thought data loss might be inherent in that case.  

I'm totally for putting attributes back on the Interface def though (named attributes or context or whatever) or doing the kwargs == attrs (even though that doesn't guarantee any implementer will actually use them).  In the kwargs case we would lose tags, but as you mentioned, we're not using them anywhere anyways. We could just drop them for all cases.  

Is the `context` option, the one everybody can agree on @econchick @matt0bi? It works for me.

tests/unit/test_plugins_loader.py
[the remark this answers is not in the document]
descent3me: hmm. i guess those expire.  it's on line 236/238

tests/unit/metrics/test_ffwd.py
descent3me: Yes, because that's what we expect the value to be. Even though the guidelines only show the `expected == actual` case, the condition is part of the expectation. We expect `metrics_sent[0]['a']` *to be equal to* `'metric'`, just like we expect `0` *to be less than* `metrics_sent[0]['value']`  (of course it reads better out loud in English as you initially suggested with (actual, expected_condition) rather than (expected_condition, actual)

gordon/plugins_loader.py
descent3me: @econchick I’m copying your comments here, but so I can address each one separately, and I've numbered them.  I'm hoping that will make it easier to discuss. My response to each comment is below the line.

>1) Hm, this forces the service operator to be aware of the entry point name that the plugin itself has declared.
---
Won't the service operator have to be aware of the entry point names for all of the plugins (since [they have to have config based on that name to be inited](https://github.com/spotify/gordon/pull/22/commits/c0b961fa349b34ff5ac361f799677b13820b7bbc#diff-41273d1d026664b64fbaab5565b7a9b5L78))?  I'm didn’t see anything that made metrics in particular different from the other plugins in this sense, so I followed the same pattern.  

> 2) Could we have the config key/section be the name without the gordon.metrics prefix, i.g. [ffwd]?
---
I don’t think there is a `gordon.metrics` section.  It should just be `metrics`, right?  I think this confusion is because we maybe picked a bad example name and used it in the naming in the tests, but there shouldn't be anything special about it, its just a string. I’ve updated it in the latest changes. Does it make more sense now?  

> 3) Then maybe add another optional [core] variable, e.g. …
```toml
[core]
plugins = ["gcp.publisher", "..."]
# or metrics_provider, or whatever
metrics = "ffwd"

[ffwd]
host = "10.99.0.1"
port = 19000
```
----
We did consider having a core variable for metrics provider name, and then a separate section for metrics configuration.  Then we decided there was no harm in passing name to the plugin, and it was nice to keep all the metrics configuration together. I think calling the section ‘metrics’, makes it more obvious what it’s for, but I don’t have too strong of an opinion and could go with ffwd/prometheus etc.  Just right now I'm sure why it would be better.

> 4) so then the code above is similar:
```python
metrics_provider = config.get('metrics')

# maybe this part is at a later commit/PR?
if not metrics_provider:  
    # instantiate & return metrics logger
if metrics_provider == 'ffwd':
    # import ffwd, instantiate w/ config, & return

for plugin_name, plugin in installed_plugins.items():
   if plugin_name == metrics_provider:
        metrics_config = config.get(metrics_provider, {})
        # load, instantiate, return
```
---
I think the first line is wrong, and I don’t really like special handling for ffwd, but overall I think it’s okay, but should be put off to a later PR when we have a default.  With these two changes I imagine it would be more like:
```
metrics_provider = config.get('core').get('metrics')
if not metrics_provider:  
    # instantiate & return metrics logger]

for plugin_name, plugin in installed_plugins.items():
   if plugin_name == metrics_provider:
        metrics_config = config.get(metrics_provider, {})
        # load, instantiate, return
```
 
> 5) This allows configuration to be abstracted away a bit better: 
---
This is depending on what we do in 3. I think your suggestion is a fine way to do it, but I’m not sure I agree the abstraction is better.  Although I like the consistency with the other plugins, I think that consistency could be confusing.  Because a metrics plugin is handled differently, it makes a little sense to configure it differently. In your example its less obvious you’re configuring something for metrics here, and could be more confused with the other plugins.  To suggest a third option, how do you feel about something like the following

```
[metrics.ffwd]
host = "10.99.0.1"
port = 19000
unit = 10E9
```
It allows it to be abstracted, but also namespaced for metrics? It would require only minimal code changes I think.

> 6) And plugin developers can still register their entry point w/ a simpler name:
---
As mentioned in 2, I think the name is not quite right. I believe it currently would be as you’re suggesting, i.e. the following.
```
# prometheus_plugin/setup.py
setup(
    # snip
    entry_points={
        'gordon.plugins': [
            ‘ffwd = path.to.callable'
        ],
    },
    # snip
)

**Comment by descent3me** [comment]
@econchick if we watching batching ability, I definitely saw it as a separate task.  It would basically be would be a new relay, and there are a few ways to do it.  One way to do it would be to subclass the relay and use a `BatchingUDPClient` (where `send` actually just stores state for the metrics until flush is called) instead. The Timer and Protocol could be reused, and only cleanup would need to be changed, but it puts all the aggregation logic in the client. 

It would probably be better write a whole new relay that stores the state, and just reuses the current Protocol and UDPClient, but only sends when cleanup is called.  In this case, the Timer may have to change, depending on if/how you want to aggregate times.

File: gordon/metrics/ffwd.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (matt0bi) [reply gordon/metrics/ffwd.py]: :+1: for the `context` option @descent3me.

File: gordon/metrics/ffwd.py (hunk 9 above)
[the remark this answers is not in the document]
Comment (matt0bi) [reply gordon/metrics/ffwd.py]: What about `SimpleFfwdRelay`? I think `FastFfwdRelay` is good. I think that an `AggregatingFfwdRelay` is a special case which needs explicit clarification whereas a non-aggregating relay is the base case. Just to be clear, I think either updating the name to `FastFfwdRelay` or adding something to the docstring would work fine here.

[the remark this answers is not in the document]
Comment (ocsw) [reply gordon/metrics/ffwd.py]: +1 to SimpleFfwdRelay and Aggregating

File: gordon/plugins_loader.py (hunk 15 above)
Comment (matt0bi) [reply gordon/plugins_loader.py]: I referenced this comment in #23. I'd be happy to update this part in as part of #23 since I think it'll be easier to address and test when there are multiple relays to pick from.

File: gordon/metrics/ffwd.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (descent3me) [reply gordon/metrics/ffwd.py]: Just to be clear, which do we like?

1) 
```
def timer(metric_name, context):
```
2)
```
def timer(metric_name, context=None):
```
3) 
```
def timer(metric_name, context=None, **kwargs):
``` 
4) 
```
def timer(metric_name, context, **kwargs):
``` 
5) 
```
def timer(metric_name, **context):
```

I was imagining 4) and I want to make sure we're on the same page.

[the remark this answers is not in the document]
Comment (matt0bi) [reply gordon/metrics/ffwd.py]: I was thinking of either 3 or 4, I think either one would work alright.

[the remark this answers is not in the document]
Comment (econchick) [reply gordon/metrics/ffwd.py]: I was thinkin 3 in case an implementation wants/needs to completely ignore it.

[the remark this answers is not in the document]
Comment (descent3me) [reply gordon/metrics/ffwd.py]: Works for me! Also, I just noticed none of these have any args descriptions, so I'll add that indicating that if used, context should be a dict.

File: gordon.toml.example (hunk 19 above)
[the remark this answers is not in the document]
Comment (descent3me) [reply gordon.toml.example]: I wouldn't mind doing it immediately after this is merged, but just because the config format still hasn't been approved yet, I don't want docs for it to be a part of this PR.

**Comment by descent3me** [comment]
@econchick @ocsw I think I've addressed all your comments. Would you mind checking back to respond or withdraw change requests as needed or let me know if I borked something else up the process?

I also don't know why [docs aren't building](https://travis-ci.org/spotify/gordon/jobs/388974081).  It seems to be complaining about [this line](https://github.com/spotify/gordon/blob/feature/add-metrics/gordon/interfaces.py#L129), but it looks correctly indented to me.

**Comment by matt0bi** [comment]
@descent3me I took a look at why sphinx is complaining and found that it doesn't like the extra indent on the ["for example" lines](https://github.com/spotify/gordon/blob/feature/add-metrics/gordon/interfaces.py#L136) in the docstrings for `incr`, `timer`, and `set`. Aligning them with the line above them makes sphinx happy again.

**Comment by descent3me** [comment]
@matt0bi Thanks!  Hmm... I really prefer them to be indented, that stinks. Maybe I'll make them shorter and try to fit on the original line.

**Comment by descent3me** [comment]
I've updated the docstrings for interfaces and metrics/ffwd.SimpleFfwdRelay. I added a link in the latter, but I'm unsure of how to find out if it will render correctly, if anybody has time to talk me through it.

File: gordon/metrics/ffwd.py (hunk 12 above)
Comment (econchick) [reply gordon/metrics/ffwd.py]: The user of ffwdtimer / other interface implementations & providers don't need to care about it implementing an interface though, right? All they should care about is the methods available to them.

File: gordon/metrics/__init__.py [hunk 20]
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2017 Spotify AB
```
econchick: 2018? :)

gordon/plugins_loader.py
econchick: I'd like to talk about this in person since I'm not exactly following this (I'll be sure to update this comment here w/ what was discussed + the outcome).

gordon/metrics/__init__.py
ocsw: Should be 2018.

ocsw:
So my remaining things are just:

1) 2017 -> 2018.
2) Remove that one unused test function and flip the two definitions after it.

gordon/metrics/__init__.py
descent3me: Yeah the whole copyright thing confuses me a bit, as far as project vs. file vs. most recent file changes, but will fix!

gordon/metrics/ffwd.py
descent3me: I guess that makes sense.  Will remove.

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,200 @@
...
+            lambda: UDPClientProtocol(message),
+            remote_addr=(self.ip, self.port))
+
+
+@zope.interface.implementer(interfaces.ITimer)
+class FfwdTimer:
+    """Timer which sends UDP messages to FFWD on completion.
+
+    Args:
+        metric (dict): Dict representation of the metric to send.
```
matt0bi: The "dict representation of the metric" makes me think that metric could be a popo (like pojo, but Python!) just to make it more defined compared to a loose dictionary. This is just a loose thought about the design, not a request to change anything btw.

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,200 @@
...
+        await self.start()
+        return self
+
+    async def __aexit__(self, *args):
+        """Exit context manager to stop timing."""
+        await self.stop()
+
+    async def start(self):
+        """Start timer."""
+        self._start_time = time.time()
```
matt0bi: Should this also be changed to `time.perf_counter` like mentioned in [here](https://github.com/spotify/gordon/pull/23#discussion_r193820381) in #23?

econchick: ah yeah I missed that - good catch @matt0bi .  Yeah I think it should be used here too.

gordon/metrics/ffwd.py
descent3me: It could be, but I think it'd be overkill for this purpose. I know there are other reasons, but generally the rule I've been using with Python is that I create an data object with attributes, it's because I know what should be in that object and want to hard fail if something tries to access it and it isn't there.  But with a [ffwd json metric](https://github.com/spotify/ffwd/tree/master/modules/json), nearly all the fields are optional, so the timer's not really counting on anything* being there, just blindly passing it through.

* you could argue we do require a 'type'='metric', but for the time being since the only thing creating timers and metrics is in the same module, I feel like it's safe enough ™.

gordon/metrics/ffwd.py
descent3me: It definitely could be.  I don't mind updating.

gordon/metrics/ffwd.py
matt0bi: Thanks, good points. I've seen code that puts things like money or IPs into objects, but those cases were closer to the core logic  for those projects. I think I've been thinking more and more about stricter checks combined with type checking, so just curious to learn more what others think around this area and where the sweet spot lies.

tests/unit/conftest.py
```diff
@@ -58,11 +59,12 @@ def __init__(self, config, success_channel, error_channel, **kwargs):
 
 @zope.interface.implementer(interfaces.IEnricherClient)
 class EnricherStub:
-    def __init__(self, config, success_channel, error_channel, **kwargs):
+    def __init__(self, config, success_channel, error_channel, metrics=None):
```
econchick: Should these still be `None`? Since we are going to provide plugins w/ a metrics object. Or actually, perhaps this part is updated in #23 ?

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,200 @@
...
+The defaults that may be overridden are shown below.
+
+.. code-block:: ini
+
+    [ffwd]
+    # to identify the service creating metrics
+    key = 'gordon-service'
+
+    # the address of the ffwd daemon (see: UDPClient)
+    ip = '127.0.0.9'
```
econchick: Nitpick, but toml syntax for strings is double quotes.

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+@zope.interface.implementer(interfaces.ITimer)
+class FfwdTimer:
+    """Timer which sends UDP messages to FFWD on completion.
+
+    Args:
+        metric (dict): Dict representation of the metric to send.
+        udp_client (UDPClient): A metric sending client.
+        time_unit (number): (optional) Scale time unit for use with
```
econchick: What do you think about using strings instead of numbers like how @matt0bi did in #23? It would make more sense as a user to say `s`, `ms`, `ns` etc.

Also, this arg's docstring should be updated to use `time.perf_counter`.

gordon/metrics/ffwd.py
```diff
@@ -0,0 +1,200 @@
...
+            'type': 'metric'
+        }
+        return metric
+
+    async def incr(self, metric_name, value=1, context=None, **kwargs):
+        """Increase a metric by 1 or a given amount.
+
+        Args:
+            metric_name (str): Identifier of the metric.
+            value (int): (optional) Value with which to increase the metric.
```
econchick: Nitpick: "default to ``1``"?

docs/config.rst
```diff
@@ -46,7 +46,12 @@ core
 
     If ``true``, Gordon will continue running even if installed & configured plugins can not be loaded. Plugin exceptions will be logged as warnings with tracebacks.
 
-    If ``false``, Gordon will exit out if it can't load one or more plugins.
+    If ``false``, Gordon will exit out if it can't load one or more plugins
```
econchick: Nitpick: I think the final `.` was removed here.

docs/config.rst
```diff
@@ -46,7 +46,12 @@ core
...
+
+.. option:: metrics='metrics-provider-name'
+
+    The metrics provider to use. Depending on the provider, more
+    configuration may be needed. See provider implementation for details.
```
econchick: No need for line breaks in actual `rst` files (unlike docstrings).

docs/config.rst
```diff
@@ -46,7 +46,12 @@ core
 
     If ``true``, Gordon will continue running even if installed & configured plugins can not be loaded. Plugin exceptions will be logged as warnings with tracebacks.
 
-    If ``false``, Gordon will exit out if it can't load one or more plugins.
+    If ``false``, Gordon will exit out if it can't load one or more plugins
+
+.. option:: metrics='metrics-provider-name'
```
econchick: It's convention to have the type rather than the "what" for the value, e.g. `.. option:: metrics="STR"` (also note double quotes).

docs/config.rst
descent3me: hmm.. will fix.

docs/config.rst
descent3me: I still think it's easier to read this way. It doesn't mess up the rendering, right?

gordon/metrics/ffwd.py
descent3me: Ah good point.

gordon/metrics/ffwd.py
descent3me: I'm not really inclined to do a string to number translation.  If it's a must have, I'll update, but I think leaving it as numbers has more flexibility, unless we add every possible time unit to the dict, which seems like too much for too little benefit.

gordon/metrics/ffwd.py
descent3me: I thought saying optional was sufficient, since if nothing is provided, it isn't scaled (i.e. it just happens we use 1 for that), but I can add it.

tests/unit/conftest.py
descent3me: Yeah, that's reasonable. Although I guess is how an implementer would do it if they don't want to care about metrics, right?

descent3me: Actually, @econchick I'll wait for your response to that ^? Since a lot of tests that don't care about metrics just broke when when I updated that, I want to make sure it's important before I go through to fix everything.

gordon/metrics/ffwd.py
descent3me: Oh sorry, still had the previous commented line in mind, when I answered. Still gonna add it, but I feel less conflicted about it now :)

gordon/metrics/ffwd.py
descent3me: I think the fact that @matt0bi did all the combinations he thought were reasonable, but you came up with another example of what is reasonable, is a good example of how this could go. I think it's perfectly reasonable that someone else might want minute resolution for example, or something else I can't think of now.  I don't think scaling factor is too complicated for the user to figure out, but as mentioned I'll add if needed.

descent3me:
Ok, @econchick I've updated about some points, and commented on the others.

tests/unit/conftest.py
econchick: Ah sorry I didn't see this earlier. I think you're right; plugins don't necessarily need to care about metrics, but they will get an object. So `metrics=None` is indeed correct, my bad.

gordon/metrics/ffwd.py
econchick: I'm still for the `s`, `ms`, `ns` etc approach, but I'd like others to weigh in on this - cc @matt0bi @ocsw @smstone

docs/config.rst
econchick: I have word wrapping in my editor, so I think it's personal preference. But if we go this route, then we should adjust the rest of the docs (which is more prevalent in `gordon-gcp` and a bit in `gordon-janitor`) to be consistent.

gordon/plugins_loader.py
```diff
@@ -227,8 +227,9 @@ def load_plugins(config, plugin_kwargs):
...
     active_plugins = _get_activated_plugins(config, installed_plugins)
     if not active_plugins:
-        return [], [], []
+        return [], [], [], None
     plugin_namespaces = _get_plugin_config_keys(active_plugins)
     plugin_configs = _load_plugin_configs(plugin_namespaces, config)
-    return _init_plugins(
+    plugin_names, plugins, errors = _init_plugins(
         active_plugins, installed_plugins, plugin_configs, plugin_kwargs)
+    return plugin_names, plugins, errors, metrics_plugin
```
econchick: perfecto, thank you!

descent3me: Ah, yeah, I actually noticed it didn't fail in my next PR. Will fix.

descent3me: Looks like a case of too much mocking...

econchick: 😬

tests/unit/metrics/test_ffwd.py
matt0bi: @ocsw @descent3me This isn't part of the feedback about this PR: It looks like there's some confusion around the `expected == actual` convention and I admit that I'm confused by it sometimes as well. Should we consider changing this guideline?

I think my confusion stems from the "_of course it reads better out loud in English_" point plus from the fact that all conditionals follow the "out loud in English" pattern (`if x > 0:` and not `if 0 < x:`).

gordon/metrics/ffwd.py
matt0bi: Hm, I like the point that "reasonable" can vary and sticking with a numeric scaling factor would keep this open and generic for those unknown unknown uses.

tests/unit/metrics/test_ffwd.py
ocsw: The logic I've heard for this boils down, IIRC, to: usually the expected value is just a variable, whereas the actual is usually a (possibly long) call, so this way you can find the expected without hunting.  I do wonder if there are other reasons; I've been told this before but it definitely can feel weird, especially with comparisons.  I don't remember which order I used for those.

ocsw: https://stackoverflow.com/questions/9331259/unit-testing-why-is-the-expected-argument-always-first-in-equality-tests

econchick: One of the goals of setting the standard and enforcing it is when your tests fail, it's clear that you know what is on each side of the equality. If we aren't consistent all the way through, I find it extremely confusing when hunting down bugs that I thought were in my code, when actually it's a typo in the expected.

Whether or not we thing expected == actual or actual == expected reads better, we choose one way and stick to it.

gordon/interfaces.py
```diff
@@ -125,14 +125,50 @@ class IPublisherClient(IGenericPlugin):
 class IMetricRelay(Interface):
     """Manage Gordon metrics."""
 
-    def incr(metric_name, value=1):
-        """Increase counter metric by 1 or a given amount."""
+    async def incr(metric_name, value=1, context=None, **kwargs):
```
econchick: Good catch on making this and `set` async methods!

PR #23, opened 2018-06-06
[WIP] Add metric logger metric relay implementation

matt0bi:

This PR implements the default metric relay, which outputs metrics to the logger.

N/A

This PR is WIP because it's based off #22. I'll update this PR with any changes in #22 that affect the code here. I'll also update this PR with any comments from #22 that are relevant to this one. 

Note that because there's an outstanding comment in #22 about the plugin loader, I didn't touch that part in this PR. Once that's resolved, I'll update the code here to tackle the "plugin loader should load logger metrics by default" issue.

**Release note**:

```release-note
Add default metric logger.
```

@spotify/alf

### DIFF
--- a/gordon/metrics/log.py
+++ b/gordon/metrics/log.py
@@ -1,0 +1,166 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Gordon ships with a default IMetricRelay implementation, which
+outputs metrics to the application logger using the standard
+library `logging` module.
+
+This implementation is used by default if no other IMetricRelay
+implementation is loaded and chosen.
+
+The default configuration that can be overriden is shown below.
+
+.. code-block:: ini
+
+    [metrics-logger]
+    # One of the log levels available from the stdlib logging module.
+    log_level = 'info'
+    # A scaling factor for timing (see: LogTimer)
+    time_unit = 1
+"""
+
+import collections
+import logging
+import time
+
+import zope.interface
+
+from gordon import interfaces
+
+# TODO: update plugin loader to load metrics logger as default
+
+
+@zope.interface.implementer(interfaces.ITimer)
+class LogTimer:
+    """Timer that outputs metrics to the logger.
+
+    Args:
+        metric (dict): Metric to output.
+        logger (.LoggerAdapter): Object used for outputting metrics.
+        time_unit (number): (optional) Scale time unit for use with
+            time.perf_counter(), for example: 1E9 to send nanoseconds.
+    """
+    def __init__(self, metric, logger, time_unit=1):
+        self.metric = metric
+        self.logger = logger
+        self.time_unit = time_unit
+
+    async def __aenter__(self):
+        await self.start()
+        return self
+
+    async def __aexit__(self, *args):
+        await self.stop()
+
+    async def start(self):
+        """Start the timer."""
+        self._start_time = time.perf_counter()
+
+    async def stop(self):
+        """Stop the timer and output the time delta to the logger."""
+        time_elapsed = (
+            time.perf_counter() - self._start_time) * self.time_unit
+        self.metric['value'] = time_elapsed
+        self.logger.log(self.metric)
+
+
+class LoggerAdapter:
+    """Configurable adapter that formats and outputs metrics.
+
+    Args:
+        level (str): Log level at which to output metric.
+    """
+    LOGFMT = '[{metric_name}] value: {value}'
+
+    def __init__(self, level):
+        self.level = level
+        self._logger = logging.getLogger('gordon.metrics-logger')
+
+    def log(self, metric):
+        """Format and output metric.
+
+        Args:
+            metric (dict): Complete metric.
+        """
+        message = self.LOGFMT.format(**metric)
+        if metric['context']:
+            message += ' context: {context}'.format(context=metric['context'])
+        getattr(self._logger, self.level.lower())(message)
+
+
+@zope.interface.implementer(interfaces.IMetricRelay)
+class LogRelay:
+    """Manage metrics that get output to a logger.
+
+    Args:
+        config (dict): Required keys:
+            level (str): Log level of metrics log messages.
+            time_unit (number): (optional) Scale time unit for use with
+                LogTimer, for example: 1E9 to send nanoseconds.
+    """
+    def __init__(self, config):
+        level = config.get('log_level', 'info')
+        self.time_unit = config.get('time_unit', 1)
+        self.logger = LoggerAdapter(level)
+        self.counters = collections.defaultdict(int)
+
+    def _create_metric(self, metric_name, value, context, **kwargs):
+        context = context or {}
+
+        return {
+            'metric_name': metric_name,
+            'value': value,
+            'context': context
+        }
+
+    async def incr(self, metric_name, value=1, context=None, **kwargs):
+        """Increase a metric counter by ``value`` and output result.
+
+        Args:
+            metric_name (str): Identifier of the metric.
+            value (number): (optional) Value with which to increase the metric.
+            context (dict): (optional) Additional key-value pairs which further
+                describe the metric, for example: {'remote-host': '1.2.3.4'}
+        """
+        self.counters[metric_name] += value
+        await self.set(metric_name, self.counters[metric_name], context,
+                       **kwargs)
+
+    def timer(self, metric_name, context=None, **kwargs):
+        """Create a LogTimer.
+
+        Args:
+            metric_name (str): Identifier of the metric.
+            context (dict): (optional) Additional key-value pairs which further
+                describe the metric, for example: {'unit': 'milliseconds'}
+        """
+        metric = self._create_metric(metric_name, None, context, **kwargs)
+        return LogTimer(metric, self.logger, self.time_unit)
+
+    async def set(self, metric_name, value, context=None, **kwargs):
+        """Output ``metric_name`` with ``value``.
+
+        Args:
+            metric_name (str): Identifier of the metric.
+            value (number): (optional) Value of the metric.
+            context (dict): (optional) Additional key-value pairs which further
+                describe the metric, for example: {'app-version': '1.5.4'}
+        """
+        metric = self._create_metric(metric_name, value, context, **kwargs)
+        self.logger.log(metric)
+
+    async def cleanup(self):
+        """Not used."""
+        pass
--- a/tests/unit/metrics/test_log.py
+++ b/tests/unit/metrics/test_log.py
@@ -1,0 +1,172 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numbers
+
+import pytest
+
+from gordon.metrics import log
+
+
+@pytest.mark.parametrize('context,context_str', [
+    ({'source': 'external'}, ' context: {\'source\': \'external\'}'),
+    ({}, '')
+])
+def test_logger_adapter_log_outputs_message(caplog, context, context_str):
+    """Metric is formatted and output to log."""
+    logger_adapter = log.LoggerAdapter('INFO')
+    metric = {
+        'metric_name': 'events',
+        'value': 12.12,
+        'context': context
+        }
+    logger_adapter.log(metric)
+    assert 1 == len(caplog.records)
+    expected_log = '[events] value: 12.12' + context_str
+    assert expected_log == caplog.records[0].msg
+
+
+@pytest.fixture
+def mock_logger_adapter(mocker):
+    return mocker.Mock(log.LoggerAdapter)
+
+
+@pytest.fixture
+def log_timer(mock_logger_adapter):
+    metric = {'metric_name': 'test-metric', 'value': 13.05, 'context': {}}
+    return log.LogTimer(metric, mock_logger_adapter, 1)
+
+
+@pytest.mark.asyncio
+async def test_log_timer_manual(log_timer):
+    """Times when using start/stop interface."""
+
+    await log_timer.start()
+    assert isinstance(log_timer._start_time, numbers.Number)
+
+    await log_timer.stop()
+    assert 1 == log_timer.logger.log.call_count
+    actual_metric = log_timer.logger.log.mock_calls[0][1][0]
+    assert 'test-metric' == actual_metric['metric_name']
+    assert {} == actual_metric['context']
+    assert 0 < actual_metric['value']
+
+
+@pytest.mark.asyncio
+async def test_log_timer_as_context_manager(log_timer):
+    """Times when using as a context manager."""
+    async with log_timer as t:
+        assert isinstance(t._start_time, numbers.Number)
+    actual_metric = log_timer.logger.log.mock_calls[0][1][0]
+    assert 1 == log_timer.logger.log.call_count
+    assert 'test-metric' == actual_metric['metric_name']
+    assert {} == actual_metric['context']
+    assert 0 < actual_metric['value']
+
+
+@pytest.fixture
+def relay_config():
+    return {
+        'time_unit': 1,
+        'log_level': 'info',
+    }
+
+
+@pytest.mark.parametrize('name,val,context', [
+    ('a-metric', 12, None),
+    ('nother-metric', 0.0003, {'source': 'internal'}),
+])
+def test_logrelay_create_metric(relay_config, name, val, context):
+    """Create a metric dictionary."""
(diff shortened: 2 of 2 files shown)


gordon/metrics/log.py
```diff
@@ -0,0 +1,123 @@
...
+        Args:
+            metric (dict): Complete metric.
+        """
+        message = self._log_fmt.format(**metric)
+        self._log_fn(message)
+
+
+@zope.interface.implementer(interfaces.IMetricRelay)
+class LogRelay:
+    """Manager metrics that get output to the logger.
```
econchick: Can I haz imperative docstring plz?

gordon/metrics/log.py
```diff
@@ -0,0 +1,123 @@
...
+
+    Args:
+        level (str): Log level at which to output metric.
+        precision (str): Used to round metric to desired number of
+            places.
+    """
+    LOGFMT = '[{metric_name}]\tvalue: {value:.<prec>f}\tattrs: {attrs}'
+
+    def __init__(self, level, precision):
+        self._log_fmt = self.LOGFMT.replace('<prec>', str(precision))
```
econchick: ooph, I wonder if there's a better way to do this 🤔

gordon/metrics/log.py
```diff
@@ -0,0 +1,123 @@
...
+@zope.interface.implementer(interfaces.IMetricRelay)
+class LogRelay:
+    """Manager metrics that get output to the logger.
+
+    Args:
+        config (dict): Required keys:
+            level (str): Log level of metrics log messages.
+            precision (int): Number of decimal places to round metric
+                value to for output.
+            time_unit (number): Scale unit to use with timers.
```
econchick: Could you add to the example config and update prose docs?

econchick: Also `time_unit` description is a little confusion, could it be similar/same as [this here](https://github.com/spotify/gordon/pull/22/commits/4e78b9e15c1870d65d56b7a12c260ceb722e4f87#diff-6c97c87bd0fada6021501e9bc1ba0925R73)?

gordon/metrics/log.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    def __init__(self, config):
+        level = config.get('log_level', 'info')
+        precision = config.get('precision', 6)
+        self.logger = MetricLogger(level, precision)
+        self.time_unit = config['time_unit']
```
econchick: Would it be reasonable to have a default one if not set?

gordon/metrics/log.py
econchick: Could this blow up if precision is anything other than an int?

gordon/metrics/log.py
```diff
@@ -0,0 +1,123 @@
...
+
+class MetricLogger:
+    """Configurable wrapper that formats and outputs metrics.
+
+    Args:
+        level (str): Log level at which to output metric.
+        precision (str): Used to round metric to desired number of
+            places.
+    """
+    LOGFMT = '[{metric_name}]\tvalue: {value:.<prec>f}\tattrs: {attrs}'
```
econchick: Any particular reason to tab-space?

econchick: Also if `attrs` are optional, maybe not have them on the format string unless it's there?

gordon/metrics/log.py
```diff
@@ -0,0 +1,123 @@
...
+        attrs = kwargs.get('attrs') or {}
+
+        return {
+            'metric_name': metric_name,
+            'value': value,
+            'attrs': attrs
+        }
+
+    async def incr(self, metric_name, value=1, **kwargs):
+        await self.set(metric_name, value, **kwargs)
```
econchick: Hmm, this wouldn't be terribly useful. Do you think there should be a stored value somewhere? i.e.

```python
async def incr(self, metric_name, value=1, **kwargs):
    self.counters['metric_name'] += value
    await self.set(metric_name, self.counters['metric_name'], **kwargs)
```


Output something like:

```
[router-msg-consumed] value: 10 attrs: ...
```

Rather than:
```
[router-msg-consumed] value: 1 attrs: ...
```

gordon/metrics/log.py
```diff
@@ -0,0 +1,123 @@
...
+    async def __aenter__(self):
+        await self.start()
+        return self
+
+    async def __aexit__(self, *args):
+        await self.stop()
+
+    async def start(self):
+        """Start the timer.
+        """
```
econchick: All one line please?

gordon/metrics/log.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+@zope.interface.implementer(interfaces.ITimer)
+class LogTimer:
+    """Timer that outputs metrics to the logger.
+
+    Args:
+        metric (dict): Metric to output.
+        logger (.MetricLogger): Object used for outputting metrics.
+        time_unit (number): (optional) Multiplier used to scale time.time()
```
econchick: I'd like some clarity here too (i.e. "`1` is seconds, `1000` for milliseconds, etc), and maybe mention that it defaults to `1`

gordon/metrics/log.py
```diff
@@ -0,0 +1,123 @@
...
+
+    async def stop(self):
+        """Stop the timer and output the time delta to the logger.
+        """
+        time_elapsed = (time.time() - self._start_time) * self.time_unit
+        self.metric['value'] = time_elapsed
+        self.logger.log(self.metric)
+
+
+class MetricLogger:
```
econchick: Is there internal magic that piggy-backs the regular logger config, i.e. sends to stream/syslog/stackdriver?

Actually, how come this isn't itself a logger (i.e. `logging.Logger`)? and registers itself as a gordon metrics logger?

gordon/metrics/log.py
matt0bi: >  ooph, I wonder if there's a better way to do this thinking

I considered combining `format` and the old `%d`-style formatting, but `replace` felt simpler and more explicit.

> Could this blow up if precision is anything other than an int?

Yep! A `ValueError` will be raised. Ideally, this value will be validated to be an int when loading the config. Also waiting for [the Great Validation](https://github.com/spotify/gordon/pull/22#discussion_r193192114).

gordon/metrics/log.py
matt0bi: Hey @econchick, I think a rebase nuked that link. Did you mean something like `Scale time unit for use with time.time() (example: 1E9 to send nanoseconds)`?

gordon/metrics/log.py
matt0bi: @econchick 

> Any particular reason to tab-space?

It's a shot at upping the readability, the image I'm going for is:

```bash
[request-acked]	value: 0.52	attrs: {'source': 'internal'}
[messages]	value: 132523	attrs: {'type': 'event'}
```

Compared to no-tabs:

```
```bash
[requests-latency] value: 0.52 attrs: {'source': 'internal'}
[messages-in-flight] value: 132523 attrs: {'type': 'event'}
```

Thoughts?

gordon/metrics/log.py
matt0bi: > Is there internal magic that piggy-backs the regular logger config, i.e. sends to stream/syslog/stackdriver?

Yeah, I think that `ulogger` sets everything up so that further calls to eg. `logging.info` make use of that set up, like it happens in [gordon/main.py](https://github.com/spotify/gordon/blob/develop/gordon/main.py#L203).

> Actually, how come this isn't itself a logger (i.e. logging.Logger)?

I don't see the benefits for the metrics logger to log to a separate logger. I don't have a strong opinion on this, so I'm curious if you think that would be a better approach?

gordon/metrics/log.py
matt0bi: :fox_face: Excellent point. I'll rework this bit to accumulate values.

gordon/metrics/log.py
descent3me: I tend to prefer the spaces.  Tabs make you think something is gonna be organized, but it usually not how it works out. I think it usually ends up coming out like the following.

```
[dnsprovider-request-sent]  value: 132523   attrs:  {'source': 'internal'}
[cpu]   value:500   attrs: None
[messages-in-flight]    value: 0.52 attrs: {'type': 'event'}
```

matt0bi: Thanks for the good example @descent3me - I got rid of the tabs.

matt0bi:
Hey @econchick, I addressed your feedback, can you PTAL again? :taco:

gordon/metrics/log.py
matt0bi: > Could you add to the example config and update prose docs?

Will do! As soon as #22 is merged and I know what's the preferred way to do this.

gordon/metrics/log.py
econchick: The benefit would be to hook into the dynamic log level changes that the introspection server plugin can give; i.e. we can turn on metrics logging even if we have ffwd (not sure if that is even possible to have two metrics relays - though an interesting idea). But more so just to toggle off/on, and to change its level. @descent3me and I even entertained the idea of having "log level"-like assignments to metric types (maybe a hack day project?).

It also seems weird to me that it's behaving like a logger, but not really as it only just makes use of `logging.<level>(msg)` and essentially attaches itself to the service logger (basically, I like separation of responsibilities). A logger gives us the ability to format how we want the message line to be ([as you're familiar with](https://github.com/spotify/ulogger/blob/master/ulogger/syslog.py#L107)), [add context](https://docs.python.org/3/howto/logging-cookbook.html#adding-contextual-information-to-your-logging-output), [add record attributes](https://docs.python.org/3/howto/logging-cookbook.html#customizing-logrecord), and [allow users to filter](https://docs.python.org/3/library/logging.html#filter-objects).

gordon/metrics/log.py
econchick: Sorry to harp on this again: `time_unit` to me translates to `ns`, `ms`, `s`, etc. Which I feel might be more common in general. Would there be any issue to switching to that?

gordon/metrics/log.py
```diff
@@ -0,0 +1,127 @@
...
+    async def __aenter__(self):
+        await self.start()
+        return self
+
+    async def __aexit__(self, *args):
+        await self.stop()
+
+    async def start(self):
+        """Start the timer."""
+        self._start_time = time.time()
```
econchick: So I recently learned that `time.time` is not guaranteed to be monotonic (ntp / clock adjustments happen often, affected by admin adjustments, etc). It'd suggest `time.perf_counter` (it's CPU time, which when we want to measure how long some logic takes, makes sense, since that differs across machines), or `time.monotonic` (less precise than perf_counter). [A good post on the differences](https://www.webucator.com/blog/2015/08/python-clocks-explained/) (disclaimer: I've only skimmed).

gordon/metrics/log.py
econchick: Here too w/r/t `ns`, `ms`, and `s`.

gordon/metrics/log.py
```diff
@@ -0,0 +1,127 @@
...
+            precision (int): Number of decimal places to round metric
+                value to for output.
+            time_unit (number): Scale unit to use with timers.
+    """
+    def __init__(self, config):
+        level = config.get('log_level', 'info')
+        precision = config.get('precision', 6)
+        self.logger = MetricLogger(level, precision)
+        self.time_unit = config.get('time_unit', 1)
+        self.counters = collections.defaultdict(int)
```
econchick: I need more defaultdicts in my life.

gordon/metrics/log.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def _create_metric(self, metric_name, value, **kwargs):
+        attrs = kwargs.get('attrs') or {}
```
econchick: I think we agreed to updated this stuff to `context` (perhaps that was after you made these changes, my apologies).

gordon/metrics/log.py
matt0bi: @econchick I'm fine with adding a bit more logic to switch between ns, ms, and s. We should apply the same approach to the ffwd module though for consistency, but maybe in the next PR to not block #22?

gordon/metrics/log.py
matt0bi: @econchick Gotcha! I still have some questions, but let's talk IRL tomorrow.

gordon/metrics/log.py
```diff
@@ -0,0 +1,173 @@
...
+    def __init__(self, metric, logger, time_unit):
+        self.metric = metric
+        self.logger = logger
+        self.time_multiplier = self._time_unit_to_multiplier(time_unit)
+
+    def _time_unit_to_multiplier(self, unit):
+        return {
+            's': 1,
+            'ms': 1E3,
+            'ns': 1E9
```
econchick: Maybe `us` for `1E6`? ha since it's 3, it could be `µs` :D

gordon/metrics/log.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def log(self, metric):
+        """Format and output metric.
+
+        Args:
+            metric (dict): Complete metric.
+        """
+        message = self.LOGFMT.format(**metric)
+        if metric['context']:
+            message += ' context: {context}'.format(context=metric['context'])
```
econchick: Would this look like:

```
[request-latency] value: 1 context: {'attr1': 'value1', 'attr2': 'value2'}
```

What do you think of expanding the context so it looks more readable/understandable:

```
[request-latency] value: 1 attr1: value1 attr2: value2
```

gordon/metrics/log.py
```diff
@@ -0,0 +1,173 @@
...
+        getattr(self._logger, self.level)(message)
```
econchick: At some point, you'll need to `lower()` the `self.level` in case it's configured in caps:

```pycon
>>> logger = logging.getLogger('')
>>> func = getattr(logger, 'INFO')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'RootLogger' object has no attribute 'INFO'
>>> func = getattr(logger, 'info')
>>>
```

gordon/metrics/log.py
```diff
@@ -0,0 +1,173 @@
...
+        context = context or {}
+
+        return {
+            'metric_name': metric_name,
+            'value': value,
+            'context': context
+        }
+
+    async def incr(self, metric_name, value=1, context=None, **kwargs):
+        """Increase a metric counter by `value` and output result.
```
econchick: `rst` formatting requires double ticks, \`\`value\`\`

gordon/metrics/log.py
```diff
@@ -0,0 +1,173 @@
...
+        Args:
+            metric_name (str): Identifier of the metric.
+            context (dict): (optional) Additional key-value pairs which further
+                describe the metric, for example: {'unit': 'milliseconds'}
+        """
+        metric = self._create_metric(metric_name, None, context, **kwargs)
+        return LogTimer(metric, self.logger, self.time_unit)
+
+    async def set(self, metric_name, value, context=None, **kwargs):
+        """Output `metric_name` with `value`.
```
econchick: Double ticks here too.

tests/unit/metrics/test_log.py
```diff
@@ -0,0 +1,172 @@
...
+
+import numbers
+
+import pytest
+
+from gordon.metrics import log
+
+
+@pytest.mark.parametrize('context,context_str', [
+    ({'source': 'external'}, ' context: {\'source\': \'external\'}'),
```
econchick: It's okay to use double quotes to avoid escaping :) but this is totally fine too.

gordon/metrics/log.py
```diff
@@ -0,0 +1,173 @@
...
+            'value': value,
+            'context': context
+        }
+
+    async def incr(self, metric_name, value=1, context=None, **kwargs):
+        """Increase a metric counter by `value` and output result.
+
+        Args:
+            metric_name (str): Identifier of the metric.
+            value (int): (optional) Value with which to increase the metric.
```
econchick: It's not really an `int` though right? Could be a float?

gordon/metrics/log.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        metric = self._create_metric(metric_name, None, context, **kwargs)
+        return LogTimer(metric, self.logger, self.time_unit)
+
+    async def set(self, metric_name, value, context=None, **kwargs):
+        """Output `metric_name` with `value`.
+
+        Args:
+            metric_name (str): Identifier of the metric.
+            value (int): (optional) Value of the metric.
```
econchick: Same here re: int v float.

tests/unit/metrics/test_log.py
```diff
@@ -0,0 +1,172 @@
...
+    assert {} == actual_metric['context']
+    assert 0 < actual_metric['value']
+
+
+@pytest.fixture
+def relay_config():
+    return {
+        'time_unit': 's',
+        'log_level': 'info',
+    }
```
econchick: What do you think about adding tests for config-related, i.e. `inof` -> Attribute error (maybe something more graceful should be thrown?), `INFO` -> does not throw an attribute error, `time_unit: fff` -> key error, etc.

gordon/metrics/log.py
econchick: Another option if more readable:
```
[request-latency] value=1 attr1=value1 attr2=value2
```

matt0bi: Just talked about this IRL, but still unsure about how to proceed.

@descent3me @smstone @ocsw @RochesterinNYC any thoughts or suggestions about formatting metric attributes/context for stdout?

gordon/metrics/log.py
matt0bi: I changed it to `number` since that's a broader type.

gordon/metrics/log.py
matt0bi: I switched this back to the "number" approach that was used before after the discussion about this in #22.

tests/unit/metrics/test_log.py
matt0bi: I'd prefer to leave this work to the config validator that's in the works.

matt0bi:
I'm going to close this PR and re-open a new one that aims to merge this branch into `develop`. 

I've addressed the most recent round of feedback - there's 1 outstanding question for others. 

> Some more nits/thoughts. <3 for the logger adaptor :D

Expensing that thesaurus was totally worth it!

gordon/metrics/log.py
descent3me: I'm not too crazy about any of 'em, and I have no better suggestions. I'd even be okay with including `context: {}` just to avoid extra formatting+concatenation.  Basically I think it's good enough as is and as readable as any of 'em, so I can live with it, but I also won't complain if you want to change it.

PR #25, opened 2018-06-11
Replace plugin interfaces with generic functional interfaces

descent3me:
- Remove IEventConsumerClient, IEnricherClient, IPublisherClient
- Add IRunnable, IMessageHandler
- Remove special message_handler logic from router
- Remove specific plugin type knowledge from router (build map dynamically)
- Allow route to be configured
- Require IEventMessags to have `phase` to route and `msg_id` for logging
- Update tests to reflect above changes

Fixes #

@spotify/alf Please review.  This commit was started to add unifying `handle_message` methods to the plugin definitions. After that change, it became apparent that specific plugins weren't needed and generic plugin definitions were sufficient to run with Gordon.

**Release note**:

```release-note
- Remove IEventConsumerClient, IEnricherClient, IPublisherClient
- Add IRunnable, IMessageHandler
- Add route configuration requirement
- Require IEventMessage to have `phase` and `msg_id`
```

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -48,7 +48,7 @@ core
 
     If ``false``, Gordon will exit out if it can't load one or more plugins.
 
-.. option:: metrics="STR"
+.. option:: metrics=STR
 
     The metrics provider to use. Depending on the provider, more
     configuration may be needed. See provider implementation for details.
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -68,3 +68,18 @@ core.logging
     .. note::
 
         If ``stackdriver`` is selected, ``ulogger[stackdriver]`` needs to be installed as its dependencies are not installed by default.
+
+
+core.route
+~~~~~~~~~~
+A table of key-value pairs of phases used to indicate the route the
+a message should take. All keys should correspond to either
+the `start_phase` attribute of a runnable plugin or the `phase` of a message
+handling plugin. Values may only correspond to `phase` of a message handling
+plugin.
+
+.. code-block:: ini
+
+    [core.route]
+    start_phase = "phase2"
+    phase2 = "phase3"
--- a/gordon/interfaces.py
+++ b/gordon/interfaces.py
@@ -25,6 +25,9 @@ class IEventMessage(Interface):
     or cleanup in case of errors.
     """
 
+    msg_id = Attribute('Identifier for the event message instance.')
+    phase = Attribute('Variable phase of the event message.')
+
     def __init__(msg_id, data, history_log, phase=None):
         """Initialize an EventMessage.
 
--- a/gordon/interfaces.py
+++ b/gordon/interfaces.py
@@ -55,14 +55,26 @@ class IEventMessage(Interface):
 class IGenericPlugin(Interface):
     """**Do not** implement this interface directly.
 
-    Use :py:class:`gordon.interfaces.IEventConsumerClient`,
-    :py:class:`gordon.interfaces.IEnricherClient`,
-    or :py:class:`gordon.interfaces.IPublisherClient` instead.
+    Use :py:class:`gordon.interfaces.IRunnable`,
+    or :py:class:`gordon.interfaces.IMessageHandler` instead.
     """
-    phase = Attribute('Plugin phase')
 
-    def __init__(config, success_channel, error_channel, metrics):
-        """Initialize an EventClient object.
+    async def shutdown():
+        """Perform any actions required to gracefully shutdown plugin."""
+
+
+class IRunnable(IGenericPlugin):
+    """Runnable plugin to produce event messages for Gordon to process.
+
+    The plugin also has the ability to send :py:class:`gordon.interfaces
+    EventMessage` objects to both success and error channels. At least
+    one runnable plugin is required to run Gordon.
+    """
+
+    start_phase = Attribute('Starting phase for event messages.')
+
+    def __init__(config, success_channel, error_channel, metrics, **kwargs):
+        """Initialize a runnable plugin.
 
         Args:
             config (dict): Plugin-specific configuration.
@@ -73,52 +85,34 @@ class IGenericPlugin(Interface):
             metrics (obj): Optional obj used to emit metrics.
         """
 
-    async def shutdown():
-        """Gracefully shutdown plugin."""
-
-
-class IEventConsumerClient(IGenericPlugin):
-    """Client for ingesting push/pull events for Gordon to process.
-
-    The client also receives both successful and failed
-    :py:class:`gordon.interfaces.IEventMessage` objects from Gordon in order
-    to perform cleanup if needed.
-    """
-
     async def run():
         """Begin consuming messages using the provided event loop."""
 
-    async def cleanup(event_msg):
-        """Perform cleanup tasks related to a message.
 
-        Args:
-            event_msg (IEventMessage): Message at the end of its life cycle.
-        """
+class IMessageHandler(IGenericPlugin):
+    """Plugin which performs some operation on an event message.
 
-
-class IEnricherClient(IGenericPlugin):
-    """Client for processing (enriching) or filtering events received by Gordon.
-
-    Note that if no extra processing is required, the implementer can
-    immediately push the received event into the success_channel.
+    The Gordon core router will use its `phase_route` to direct messages
+    produced by any runnable plugins the appropriate message handling
+    plugins, identified by their phase attribute. At least
+    one message handling plugin is required to run Gordon.
     """
 
-    async def process(event_msg):
-        """Process message.
+    phase = Attribute('Plugin phase')
+
+    def __init__(config, metrics, **kwargs):
+        """Initialize a message handler.
 
         Args:
-            event_msg (IEventMessage): Message to process.
+            config (dict): Plugin-specific configuration.
+            metrics (obj): Obj used to emit metrics.
         """
 
-
-class IPublisherClient(IGenericPlugin):
-    """Client for publishing processed events to their destination."""
-
-    async def publish_changes(event_msg):
-        """Publish processed event to its destination.
+    async def handle_message(event_message):
+        """Perform some operation on or triggered by an event message.
 
         Args:
-            event_msg (IEventMessage): Message ready to be sent to destination.
+            event_message (IEventMessage): Message on which to operate.
         """
 
 
--- a/gordon/main.py
+++ b/gordon/main.py
@@ -139,8 +139,9 @@ def _gather_plugins_by_type(plugins, debug):
     return runnable_plugins, message_handling_plugins
 
 
-def _setup_message_router(plugins, success_channel, error_channel):
-    msg_router = router.GordonRouter(success_channel, error_channel, plugins)
+def _setup_router(config, plugins, success_channel, error_channel):
+    msg_router = router.GordonRouter(
+        config, success_channel, error_channel, plugins)
     return msg_router
 
 
@@ -174,7 +175,9 @@ def run(config_root):
         logging.info(f'Loaded {len(plugin_names)} plugins: {plugin_names}.')
 
     runnables, message_handlers = _gather_plugins_by_type(plugins, debug_mode)
-    msg_router = _setup_message_router(message_handlers, **channels)
+
+    route_config = config.get('core', {}).get('route', {})
+    msg_router = _setup_router(route_config, message_handlers, **channels)
 
     logging.info(f'Starting gordon v{version}...')
     loop = asyncio.get_event_loop()
--- a/gordon/main.py
+++ b/gordon/main.py
@@ -102,75 +102,50 @@ def _log_or_exit_on_exceptions(base_msg, exc, debug):
         raise SystemExit(1)
 
 
-def _assert_required_plugins(installed_plugins, debug):
-    # enricher not required
-    required_providers_available = {
-        'event_consumer': False,
-        'publisher': False,
-    }
-    for plugin in installed_plugins:
-        if interfaces.IEventConsumerClient.providedBy(plugin):
-            required_providers_available['event_consumer'] = True
-        elif interfaces.IPublisherClient.providedBy(plugin):
-            required_providers_available['publisher'] = True
-
-    missing = []
-    msg = ('The provider for the "{name}" interface is not configured for the '
-           'Gordon service or is not implemented.')
-    for provider, available in required_providers_available.items():
-        if not available:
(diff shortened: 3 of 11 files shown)


gordon/interfaces.py
```diff
@@ -53,16 +53,29 @@ def update_phase(new_phase):
 
 
 class IGenericPlugin(Interface):
-    """**Do not** implement this interface directly.
+    """Plugin which is associated with a phase.
```
econchick: Actually, you can remove the `phase` attribute here, it's not needed for runnable plugins.

gordon/interfaces.py
```diff
@@ -53,16 +53,29 @@ def update_phase(new_phase):
...
+class IRunnable(IGenericPlugin):
+    """Plugin for ingesting push/pull events for Gordon to process.
+
+    The plugin also has the ability to send :py:class:`gordon.interfaces
+    EventMessage` objects to both success and error channels. At least
+    one runnable plugin is required to run Gordon.
+    """
+
+    def __init__(config, success_channel, error_channel, metrics, **kwargs):
+        """Initialize an runnable.
```
econchick: `an` -> `a`; also  probably want to say a runnable _what_.

gordon/main.py
```diff
@@ -148,28 +124,21 @@ def _gather_runnable_plugins(plugins, debug):
     return plugins_to_run
```
econchick: Can't comment on the lines directly above, but I _think_ the `elif` branch can be removed now, right?

tests/unit/conftest.py
```diff
@@ -28,66 +28,63 @@
 from gordon import interfaces
 
 
-PLUGIN_NAMES = ['event_consumer', 'enricher', 'publisher', 'runnable']
+PLUGIN_NAMES = [
+    'event_consumer',
+    'enricher',
+    'publisher',
+    'cleaner',
```
econchick: I know you don't like the idea of the event consumer doing the cleaning up, and I'm not sure the interface definition technically separates that out, but I'd _really_ like to talk that part through before we start doing it, even eluding to it in tests. The main reason is (unfortunately) the gcp event consumer itself is closely tied with the cleanup phase as the consuming of a message is closely integrated with the ack'ing/cleaning up of the message. We'd need to think about how to untangle that first.

gordon/interfaces.py
descent3me: Yeah, this attribute is the main piece that made me split the cleaner out. I initially was thinking that that the event consumer would implement both runnable and message handler.

gordon/interfaces.py
descent3me: Will fix!

gordon/main.py
descent3me: The elif checks if something implements runnable without explicitly declaring that it implements runnable.  I'm fine with dropping it!

tests/unit/conftest.py
descent3me: The reason I broke it out was because it was breaking the phase attribute defined in the generic plugin interface and complicated the routing logic a bit.

gordon/main.py
econchick: yeah I think I'd be fine w/ dropping it too.

econchick:
@descent3me totally fair to spend some time investigating the implications; curious to see what you find!

descent3me:
@econchick What I made while experimenting is the on this [wip branch](https://github.com/spotify/gordon-gcp/compare/wip/test_event_cleaner).

I realize there may totally be important consequences that I'm missing, so please have a look. The implementation is not that different, just logic moved around slightly so the plugins could be separated.  The only thing that concerns me is [this comment](https://github.com/spotify/gordon-gcp/blob/develop/tests/unit/plugins/service/test_gdns_publisher.py#L289), but as it existed before my change, on the surface it doesn't look worse than the previous implementation.

docs/config.rst
```diff
@@ -48,6 +48,11 @@ core
 
     If ``false``, Gordon will exit out if it can't load one or more plugins.
 
+.. option:: route=DICT-OF-STRINGS
```
descent3me: Totally guessing here. I wasn't able to find a page that described how to format these by googling alone.  Anybody have a reference to recommend?

gordon/interfaces.py
econchick: I apologize for my short term memory since we _just_ talked about this yesterday 🤣 but does the `phase` attribute need to live here? It doesn't make sense to require a runnable to have a phase.

gordon/interfaces.py
```diff
@@ -73,52 +88,26 @@ def __init__(config, success_channel, error_channel, metrics=None):
...
-    """Client for publishing processed events to their destination."""
-
-    async def publish_changes(event_msg):
-        """Publish processed event to its destination.
+    async def handle_message(event_message):
+        """Perform some operation on or triggered by an event message.
 
         Args:
-            event_msg (IEventMessage): Message ready to be sent to destination.
+            event_message (IEventMessage): Message on which to operate.
```
econchick: <3 not ending on a preposition, hah

gordon/interfaces.py
```diff
@@ -73,52 +88,26 @@ def __init__(config, success_channel, error_channel, metrics=None):
...
 
-    async def cleanup(event_msg):
-        """Perform cleanup tasks related to a message.
-
-        Args:
-            event_msg (IEventMessage): Message at the end of its life cycle.
-        """
 
+class IMessageHandler(IGenericPlugin):
+    """Plugin which performs some operation on an event message."""
```
econchick: I wonder if it's worth mentioning that the router within core will route messages from/to whomever implements a message handler. I think the sentence you have here is perfect, so maybe start another line (like how the old IEventConsumerClient had).

gordon/interfaces.py
```diff
@@ -73,52 +88,26 @@ def __init__(config, success_channel, error_channel, metrics=None):
...
-class IEnricherClient(IGenericPlugin):
-    """Client for processing (enriching) or filtering events received by Gordon.
-
-    Note that if no extra processing is required, the implementer can
-    immediately push the received event into the success_channel.
-    """
-
-    async def process(event_msg):
-        """Process message.
+    def __init__(config, metrics, **kwargs):
```
econchick: The "event consumer"-type message handler is a special case, which should get the success channel, since we're expecting something to consume from an external queue to produce messages for the internal queue.

One way is to have a special interface that inherits from both this and runnable. I'm not necessarily married to that. But perhaps _for now_ it'd be enough to just make it clear in prose docs that we expect an implementation that implements both runnable and message handler in order to produce messages for the router to then route.

gordon/main.py
```diff
@@ -102,74 +102,41 @@ def _log_or_exit_on_exceptions(base_msg, exc, debug):
...
-        if interfaces.IEventConsumerClient.providedBy(plugin):
-            implemented['event_consumer'] = plugin
-        elif interfaces.IEnricherClient.providedBy(plugin):
-            implemented['enricher'] = plugin
-        elif interfaces.IPublisherClient.providedBy(plugin):
-            implemented['publisher'] = plugin
-    return implemented
+def _assert_required_plugins(plugins, debug):
+    if not plugins:
+        msg = (f'At least one runnable plugin is required.')
```
econchick: Should we require only a runnable one? or also one that implements a message handler so that it can be routed somewhere? The begs the question of "should we force plugin implementers/operators to have more than one plugin" since while I can foresee a use case (that I can't think of for now) that only uses one plugin, I'm not sure if the router would work without at least two it its current state.

docs/config.rst
```diff
@@ -48,6 +48,11 @@ core
...
+    The route message should take through Gordon, defined via the phases each
+    plugin implements. Route should be a dict key-value pairs of
+    ``{'from-phase': 'to-phase'}``.
```
econchick: This is helpful, but the example should probably be in `toml` format, like you have added in the example:

 ```toml
[core]
plugins = ["one.plugin", "two.plugin"]
debug = false
[core.route]
from_phase1 = "to_phase2"
from_phase2 = "to_phase3"
```

So then it should have a separate heading like `[core.logging]` below, and therefore the option documentation is just key="value" sort of thing.

File: gordon/main.py [hunk 11]
```diff
@@ -166,7 +167,8 @@ def run(config_root):
     if plugin_names:
         logging.info(f'Loaded {len(plugin_names)} plugins: {plugin_names}.')
 
-    msg_router = _setup_message_router(plugins, **channels)
+    route_config = config.get('core', {}).get('route', {})
```
econchick: How you have the `GordonRouter` initialization right now expects this to be set. Should we set a default? IMO it might be a good idea, like the general happy path as default.

docs/config.rst
descent3me: I started with that, but then wasn't sure what to do name the .. option part, since the keys can be whatever.  Should I put something like `.. option:: from_phase=STR`? Or is there another way to indicate the keys are arbitrary?

gordon/interfaces.py
descent3me: Ah, good point. I just forgot to move it. Will fix!

docs/config.rst
econchick: I'm not aware of the best way to do that; I don't know of any project off the top of my head that allows arbitrary keys. But I think what you've proposed is what I would have done, noticing that multiple are allowed. I wonder if it's worth saying keys should be unique (for now)?

gordon/interfaces.py
descent3me: I thought about this, too, especially since I didn't like the conflicting init methods, but wasn't sure since we're not verifying yet. I'm fine with adding it, though.

gordon/main.py
descent3me: True, in that case it's not that it requires a message handler, but it requires a message handler that has phase 'cleanup', since that's where it would buckle. I'd almost rather just change the router, either to make final stages optional or have a default drop functionality for 'cleanup'. Alternatively, I could just add a check for 'cleanup' here. I'm slightly against requiring specific phases, so I think I like having a default for 'cleanup' best. Thoughts?

gordon/main.py
descent3me: I liked `.* -> 'cleanup' (with default to drop message)` as the default, which because of your above comment I'm now realizing doesn't work like I thought it did (for whatever we go with, I will add a test this time!). 

I'm less inclined to put a default happy path in the router itself, just because that seems very implementation-specific. If we do, which would you want `consume->enrich->publish->cleanup` or `consume->publish->cleanup`? And should configured routes merge with the default or do a full overwrite?

econchick: I think both could be the default path, and that it could be deduced if there's an `enrich` phase plugin, much like how we have currently. What does @matt0bi think about default path(s)?

gordon/router.py
```diff
@@ -103,8 +103,8 @@ def _get_next_phase(self, event_msg):
         event_msg.update_phase(next_phase)
         next_plugin = self.phase_plugin_map.get(next_phase)
         if next_phase in self.FINAL_PHASES and not next_plugin:
-            msg = (f'Dropping message"{event_msg}", final phase "{next_phase}"'
-                   'not implemented.')
+            msg = (f'Dropping message"{event_msg.msg_id}", final phase '
+                   f'"{next_phase}" not implemented.')
```
descent3me: When adding this line in an earlier commit, I noticed that what we were outputting might not be that useful, and we hadn't required that EventMessages have the phase attribute that the router uses either, which could cause Gordon to crash, so I added this commit.

descent3me:
@econchick I have addressed many of the comments and updated the PR, if you have time to review again.

gordon/interfaces.py
```diff
@@ -25,6 +25,9 @@ class IEventMessage(Interface):
     or cleanup in case of errors.
     """
 
+    msg_id = Attribute('Identifier for the event message instance.')
```
econchick: Is there added value to having this defined here rather than in the `__init__` below? As I understood it, one can define required attributes like this way, or in the `__init__` method, that it's redundant to do both. I could definitely have the wrong understanding; zope docs are lacking in this area. But I think it should be one or the other, not both.

What I didn't find clear in the zope docs was if defining attributes up here was defining class attributes, rather than instance attributes. Their examples show otherwise though, so I don't think these `Attribute()`s meant strictly for class attributes. But to avoid confusion, @matt0bi and I put these in the `__init__`.

🤔 Did you find that these failed verification otherwise?

gordon/router.py
econchick: Nice catch! Would you mind adding a space between `message` and `"{event_msg.msg_id}"`? That was my mistake originally, didn't even notice it before.

(sidebar topic) What I think I intended when actually writing this string message was to define a `__repr__` method (which would get implicitly called here) that would go something like

```python
# implements IEventMessage
class SomeEventMessage:
    # snip
    def __repr__(self):
        return f'<SomeEventMessage(id={self.msg_id})>'
```

and therefore would look like:

```
'Dropping message "<SomeEventMessage(id=1234)>"...'
```

I'm not making this suggestion for here; I just would like to start using `__repr__`s more to our advantage, since we can stuff it with as much (or as little) as we want, and just need to use `{<instance_name>}` rather than `{<instance_name.some_attribute}`.

gordon/interfaces.py
descent3me: Actually, now that I wrote it I kind of like saving it as `msg.id` better, but maybe that's just confusing given the init parameter.  I don't have any strong feelings about removing it from the init declaration. It gives the opportunity for the `id` to be determined by something that isn't passed into the init, but it's not that big of a loss for me.

econchick: I didn't think of it that way - wish I had initially since it's a very good point. You've convinced me to go with `Attributes` like you have here, and that having it in both is fine (but may or may not be what one wants, depending on whatever the use case is). I would definitely shy away from naming anything `id` since it's a reserved keyword in Python.

I'd actually be okay with removing it (and `phase`) from `__init__`, if you do want to/isn't a lot of work, since we don't need to force a guarantee that whomever produces these messages provides upon initialization, just that those attributes are there.

Would there be harm in moving the rest of the init attributes up outside of `__init__`? I can't immediately think of any 🤔

descent3me: Phase could be removed from init, but I think I wouldn't want to require data and history_log as attributes, since we have only been using those as private variables (i.e. nothing else should access them directly at the core level, and should instead call `update_history` (and maybe later when we want it `dump_history` (or better name)) and leave the implementation of how they want to store it to the implementer.  However in the name of getting this merged in order to update gordon-gcp, I'm going to leave it as-is for now, but I will note it down for potential optimizations.

PR #27, opened 2018-06-13
Add metrics instrumentation to router

econchick:

This gives default metrics, no matter which plugins are used/installed. Because of that, the metric names are prefixed with `router-`. Metrics added:

- Messages consumed (to be routed if valid interface provider)
- Messages dropped (either invalid message interface provider, or error’ed out)
- Messages in flight (a gauge that’s updated each time a message is routed)
- Duration of messages in flight (consume->cleanup)
- Messages successfully completed route from start to finish (consume -> publish)
- Messages routed per phase (i.e. from consume to publish, from enrich to publish, etc)

Fixes #

Depending on which gets in first, either this or #25 will need to rebase.

**Release note**:

```release-note
Add basic metrics to core router.
```

### DIFF
--- a/gordon/router.py
+++ b/gordon/router.py
@@ -58,16 +58,21 @@ class GordonRouter:
             :class:`gordon.interfaces.IEventMessage` s that were not
             processed due to problems.
         plugins (list): Instantiated message handling plugins.
+        metrics (obj): Implemented :class:`IMetricRelay` interface
+            to emit metrics.
     """
     # TODO (lynn): Ideally this is configurable/extendable in future
     #              iterations of gordon.
     FINAL_PHASES = ('cleanup',)
 
-    def __init__(self, phase_route, success_channel, error_channel, plugins):
+    def __init__(self, phase_route, success_channel, error_channel, plugins,
+                 metrics):
         self.success_channel = success_channel
         self.error_channel = error_channel
         self.phase_plugin_map = self._get_phase_plugin_map(plugins)
         self.phase_route = phase_route
+        self.metrics = metrics
+        self._messages_in_flight = {}
 
     def _get_phase_plugin_map(self, plugins):
         phase_map = {p.phase: p for p in plugins}
@@ -90,20 +95,44 @@ class GordonRouter:
             next_phase = 'cleanup'
         return next_phase
 
-    async def _route(self, event_msg, next_phase=None):
-        if not interfaces.IEventMessage.providedBy(event_msg):
-            msg = (f'Ignoring message "{event_msg.msg_id}". Does not correctly'
-                   ' implement `IEventMessage`.')
-            logging.warn(msg)
-            return
+    async def _update_messages_in_flight(self):
+        in_flight_count = len(self._messages_in_flight)
+        await self.metrics.set(
+            'router-messages-in-flight', value=in_flight_count)
 
-        if not next_phase:
+    async def _add_message_in_flight(self, event_msg):
+        if event_msg.msg_id not in self._messages_in_flight:
+            context = {'unit': 'seconds'}
+            timer = self.metrics.timer(
+                'router-message-flight-duration', context=context)
+            await timer.start()
+            self._messages_in_flight[event_msg.msg_id] = timer
+
+        await self._update_messages_in_flight()
+
+    async def _remove_message_in_flight(self, event_msg):
+        if event_msg.phase in self.FINAL_PHASES:
+            timer = self._messages_in_flight.pop(event_msg.msg_id)
+            await timer.stop()
+
+        await self._update_messages_in_flight()
+
+    async def _route(self, event_msg, force_cleanup=False):
+        if force_cleanup:
+            next_phase = 'cleanup'
+        else:
             next_phase = self._get_next_phase(event_msg)
+
+        context = {
+            'current': event_msg.phase,
+            'next': next_phase,
+        }
+        await self.metrics.incr('router-update-message-phase', context=context)
 
         event_msg.update_phase(next_phase)
         next_plugin = self.phase_plugin_map.get(next_phase)
         if next_phase in self.FINAL_PHASES and not next_plugin:
-            msg = (f'Dropping message"{event_msg.msg_id}", final phase '
+            msg = (f'Dropping message "{event_msg.msg_id}", final phase '
                    f'"{next_phase}" not implemented.')
             logging.debug(msg)
             return
@@ -115,13 +144,30 @@ class GordonRouter:
                 msg = f'Adding message "{event_msg}" to success channel.'
                 event_msg.append_to_history(msg, event_msg.phase)
                 await self.success_channel.put(event_msg)
+            elif next_phase in self.FINAL_PHASES and not force_cleanup:
+                # if we're here, can assume message successfully went thru
+                # the entire phase route and not dropped
+                await self.metrics.incr('router-message-completed')
 
         except Exception as e:
             msg = (f'Routing message "{event_msg}" to cleanup due to exception:'
                    f' "{e}"')
             event_msg.append_to_history(msg, event_msg.phase)
             logging.warn(msg, exc_info=e)
-            await self._route(event_msg, 'cleanup')
+            context = {'error': e.__class__.__name__}
+            await self.metrics.incr('router-message-dropped', context=context)
+            await self._route(event_msg, force_cleanup=True)
+
+    async def _verify_message_impl(self, event_msg):
+        if not interfaces.IEventMessage.providedBy(event_msg):
+            msg = (f'Ignoring message "{event_msg.msg_id}". Does not correctly'
+                   ' implement `IEventMessage`.')
+            logging.warn(msg)
+
+            context = {'error': 'invalid-message-provider'}
+            await self.metrics.incr('router-message-dropped', context=context)
+            return False
+        return True
 
     async def _poll_channel(self):
         while True:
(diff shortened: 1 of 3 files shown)


econchick:
whoops - forgot the cc @spotify/alf 👋

gordon/router.py
```diff
@@ -142,13 +172,22 @@ def _get_next_phase(self, event_msg):
...
         except Exception as e:
             msg = f'Dropping message "{event_msg}" due to exception: "{e}"'
             event_msg.append_to_history(msg, event_msg.phase)
             logging.warn(msg, exc_info=e)
+            context = {'error': e.__class__.__name__}
+            await self.metrics.incr('router-message-dropped', context=context)
             await self._route(event_msg, 'cleanup')
 
+        finally:
+            await self._update_messages_in_flight(event_msg)
```
smstone: you're calling `await self._update_messages_in_flight(event_msg)` twice in this function. is that needed?

gordon/router.py
```diff
@@ -121,16 +123,44 @@ def _get_next_phase(self, event_msg):
...
+            context = {'error': 'invalid-message-provider'}
+            await self.metrics.incr('router-message-dropped', context=context)
             return
 
+        await self._update_messages_in_flight(event_msg)
+
         if not next_phase:
             next_phase = self._get_next_phase(event_msg)
 
+        old_phase = event_msg.phase
```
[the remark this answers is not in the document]
econchick: I thought about that, but I want to use `old_phase` in line 175 (where I see if the message originally came from the `publish` phase). I thought that comparison would be misleading. Do you find it more misleading here and less so down there to use `current_phase`?

gordon/router.py
```diff
@@ -121,16 +123,44 @@ def _get_next_phase(self, event_msg):
...
+        context = {
+            'current': old_phase,
+            'next': next_phase,
+        }
+        await self.metrics.incr('router-update-message-phase', context=context)
```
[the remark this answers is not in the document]
econchick: Ah I missed that one in the commit - good catch. This is to see message by phase. @descent3me and I were talking through what to measure last week, and I originally had this idea for each plugin, i.e. "enricher initially receives message" and "publisher initially receives message" etc, so we can track the difference between consume -> publish and consume -> enrich -> publish. But then @descent3me noticed that this can just live in the router since it knows all this info already.

gordon/router.py
```diff
@@ -142,13 +172,22 @@ def _get_next_phase(self, event_msg):
                 msg = f'Adding message "{event_msg}" to success channel.'
                 event_msg.append_to_history(msg, event_msg.phase)
                 await self.success_channel.put(event_msg)
+            elif old_phase == 'publish':
```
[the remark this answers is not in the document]
econchick: (love the word penultimate)

Paging @descent3me to this convo since she & I have worked/talked about the phase route. As it stands, there are fixed phases that Gordon supports, although we want to eventually support multiple phases/paths, we're limiting it now for mvp. So yes as it stands, `publish` is designated as the final phase before cleaning up.

gordon/router.py
```diff
@@ -142,13 +172,22 @@ def _get_next_phase(self, event_msg):
...
+                # if we're here, can assume message successfully went thru
+                # the entire phase route
```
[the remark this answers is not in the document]
econchick: If the message is at this point, that means it originally came from `publish`, it successfully went through the  `cleanup` phase. Because the only two valid happy paths are `consume` -> `enrich` -> `publish` -> `cleanup` and `consume` -> `publish` -> `cleanup`, it is at this point that we can assume it never got dropped for some error, and it successfully completed the designated path. Does that make sense?

gordon/router.py
econchick: Yes it was intentional. The first time ensures that the message is in memory (if we've already seen it, then it does nothing except send the gauge metric, which is fine). And this second time to ensure that if a message is dropped or done, it gets removed from memory. All to ensure we give an accurate count of messages in flight.

Perhaps there's a better way to do this?

gordon/router.py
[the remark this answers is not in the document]
econchick: (updated the PR message, will update the actual long-desc of commit message too)

gordon/router.py
```diff
@@ -121,16 +123,44 @@ def _get_next_phase(self, event_msg):
...
+            context = {'unit': 'seconds'}
+            timer = self.metrics.timer(
+                'router-message-flight-duration', context=context)
+            await timer.start()
+            self._messages_in_flight[event_msg.msg_id] = timer
+
+        elif event_msg.phase in self.FINAL_PHASES:
+            timer = self._messages_in_flight.pop(event_msg.msg_id)
+            await timer.stop()
+
```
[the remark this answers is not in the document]
econchick: oh sorry, should have clarified what you meant: do you mean add, or remove this blank line? I have this current extra line here to make it more readable.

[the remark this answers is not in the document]
smstone: ah ok, never mind!

gordon/router.py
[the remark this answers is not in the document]
econchick: @descent3me what happens if `next_phase` is `cleanup` and `old_phase` is `consume`? I wouldn't want to count that by accident; just limiting it to the full happy path.

gordon/router.py
descent3me: It seems like there's two things in `_update_messages_in_flight`. Maybe could split out an ` _add_to_messages_in_flight` function?

econchick:
@smstone @descent3me I reworked this a little bit to hopefully make more sense. Mind a re-review?

gordon/router.py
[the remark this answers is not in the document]
descent3me: Aha, I see what you're trying to measure now. For the metric I was thinking of I would expect it to get emitted even if old_phase if 'consume', because it still shows the message was routed through successfully.

What you are describing sounds like a good metric to track, but it also sounds very implementation-specific. I would expect to track that in the plugin itself rather than the router (i.e. the GDNSPublisher emits a 'successfully-published' metric).

gordon/router.py
```diff
@@ -90,20 +95,45 @@ def _get_next_phase(self, event_msg):
...
+        if event_msg.phase in self.FINAL_PHASES:
+            timer = self._messages_in_flight.pop(event_msg.msg_id)
+            await timer.stop()
+
+        await self._update_messages_in_flight()
+
+    async def _route(self, event_msg, force_cleanup=False):
+        if force_cleanup:
+            next_phase = 'cleanup'
+        else:
```
econchick: I know this boolean check could be handled a bit cleaner, but I wanted to be super explicit.

econchick:
hey @smstone @descent3me - mind a re-review when you have some time?

tests/unit/test_router.py
```diff
@@ -99,6 +147,57 @@ def test_get_next_phase(current_phase, exp_next_phase, router_inst, event_msg,
     assert 1 == len(caplog.records)
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize('exists,exp_start_count', (
+    (True, 0),
+    (False, 1),
+))
+async def test_add_message_in_flight(exists, exp_start_count, router_inst,
```
descent3me: I'm a little unclear on what `exists` is here.  Is it if the next plugin exists?

gordon/router.py
```diff
@@ -90,20 +95,45 @@ def _get_next_phase(self, event_msg):
...
             next_phase = self._get_next_phase(event_msg)
 
+        old_phase = event_msg.phase
```
descent3me: We could probably drop this variable now.

econchick: dangit I meant to!

tests/unit/test_router.py
econchick: It's actually if the event_msg exists (in the dict already). Would `msg_exists` work?

gordon/router.py
```diff
@@ -135,7 +181,13 @@ def _get_next_phase(self, event_msg):
                 #              to clean up rather than just breaking
                 break
 
+            if not await self._verify_message_impl(event_msg):
+                break
+
+            await self.metrics.incr('router-message-consumed')
+            await self._add_message_in_flight(event_msg)
             await self._route(event_msg)
+            await self._remove_message_in_flight(event_msg)
```
smstone: Just for my clarification, why are we removing messages from being "in flight" after they are routed? Aren't messages pretty much in flight the entire time unless dropped or completed?

smstone:
LGTM, just one clarifying question

gordon/router.py
econchick: Sure - we're removing from the in-memory dictionary so we can send an accurate count of messages in flight when we "set" the gauge metric.

descent3me: Maybe the confusion is from the method name (but I think `remove_message_in_flight_if_appropriate` is a bit long)? If you look in the method definition they are only removed after their final routing, just like they're only added before their first routing.

tests/unit/test_router.py
```diff
@@ -99,6 +147,57 @@ def test_get_next_phase(current_phase, exp_next_phase, router_inst, event_msg,
     assert 1 == len(caplog.records)
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize('msg_exists,exp_start_count', (
+    (True, 0),
+    (False, 1),
+))
+async def test_add_message_in_flight(msg_exists, exp_start_count, router_inst,
```
descent3me: 👍

PR #28, opened 2018-06-13
Add metric logger MetricRelay implementation

matt0bi:

This PR implements the default metric relay, which outputs metrics to a logger.

Fixes #

This PR has addressed all the feedback from #23 with [one piece of feedback](https://github.com/spotify/gordon/pull/23#discussion_r194443794) waiting for more input/opinions. 

This PR does not include the changes needed to make the metrics logger the default logger.

**Release note**:

```release-note
Add metric logger.
```

@spotify/alf

### DIFF
--- a/gordon/metrics/log.py
+++ b/gordon/metrics/log.py
@@ -1,0 +1,164 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Gordon ships with a default IMetricRelay implementation, which
+outputs metrics to the application logger using the standard
+library `logging` module.
+
+This implementation is used by default if no other IMetricRelay
+implementation is loaded and chosen.
+
+The default configuration that can be overriden is shown below.
+
+.. code-block:: ini
+
+    [metrics-logger]
+    # One of the log levels available from the stdlib logging module.
+    log_level = 'info'
+    # A scaling factor for timing (see: LogTimer)
+    time_unit = 1
+"""
+
+import collections
+import logging
+import time
+
+import zope.interface
+
+from gordon import interfaces
+
+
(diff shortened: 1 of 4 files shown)


gordon/metrics/log.py
```diff
@@ -0,0 +1,164 @@
...
+    def log(self, metric):
+        """Format and output metric.
+
+        Args:
+            metric (dict): Complete metric.
+        """
+        message = self.LOGFMT.format(**metric)
+        if metric['context']:
+            message += ' context: {context}'.format(context=metric['context'])
+        getattr(self._logger, self.level.lower())(message)
```
descent3me: instead of calling `lower()` every message, should we just do it once on initialization?

descent3me: Actually should we call `upper()` on initialization and use [log()](https://docs.python.org/3/library/logging.html#logging.Logger.log) instead?  I think it's a bit clearer than `getattr`.

econchick: I agree that `lower` should be called once (and maybe `getattr` once on instantiation too), but I'm not sure I'm understanding how that `log` method works?

```pycon
>>> import logging
>>> logger = logging.getLogger('')
>>> logger.log('INFO', 'hi')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/lynn/.pyenv/versions/3.6.2/lib/python3.6/logging/__init__.py", line 1368, in log
    raise TypeError("level must be an integer")
TypeError: level must be an integer
>>>
# same with 'info'
```

descent3me: AFAICT it works the same as `info` or `debug`, et. al., but you have to pass the level constant in as the first argument.
```
>>> import logging
>>> logger = logging.getLogger('')
>>> logger.log(logging.INFO, 'hi')
```
So I initially wrote to do `upper()` in the `__init__` so that you could have like 
`self.level = getattr(logging, level.upper())`
and then `level` would be the `logging.LEVEL` rather than a string. 

But then I thought (and edited the comment) why not have the `__init__` take `logging.LEVEL` instead. I guess it's just moving the config -> log level parsing logic from the adapter to the relay, just cuz I thought it fit a little better there, but I could go either way on it.

econchick:
@descent3me mind a re-review?

tests/unit/metrics/test_log.py
```diff
@@ -0,0 +1,172 @@
...
+    expected = {
+        'metric_name': name,
+        'value': val,
+        'context': context or {}
+    }
+    assert expected == actual
+
+
+@pytest.fixture
+def mock_log_relay(monkeypatch, mocker, relay_config):
```
descent3me: Could we rename it something like `log_relay` or `log_relay_inst`? Prefacing it with `mock` makes me think the relay itself is mocked out.

econchick:
@descent3me done! lmk if you find something else

PR #30, opened 2018-06-15
Add basic graceful shutdown

econchick:

Adds a basic graceful shutdown hook when the service receives a `SIGINT`, `SIGTERM`, or a `SIGHUP` signal. All it does is just cancels the outstanding asyncio tasks.

Fixes #

For where I arrived at this approach, I found [this SO answer](https://stackoverflow.com/a/43810272) on awaiting cancelled tasks (and not just cancelling them), used [this example approach](https://gist.github.com/nvgoldin/30cea3c04ee0796ebd0489aa62bcf00a) combined [with this one](https://stackoverflow.com/questions/23313720/asyncio-how-can-coroutines-be-used-in-signal-handlers).

**Release note**:

```release-note
Add a basic graceful shutdown mechanism.
```

cc @spotify/alf

### DIFF
--- a/gordon/main.py
+++ b/gordon/main.py
@@ -35,6 +35,7 @@ Example:
 import asyncio
 import logging
 import os
+import signal
 
 import click
 import toml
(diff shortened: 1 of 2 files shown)


descent3me:
lgtm +1

econchick:
🙏 thanks!

PR #32, opened 2018-06-18
Fix args for seting up router

econchick:

I noticed this issue when trying to help get Gordon to run. Totally open to a better/more elegant way to do this.

Fixes #

FYI another release for this package will need to be made.

**Release note**:

```release-note
Fix router setup.
```

cc @spotify/alf

### DIFF
--- a/gordon/main.py
+++ b/gordon/main.py
@@ -177,12 +177,12 @@ def run(config_root):
     config = setup(os.path.abspath(config_root))
     debug_mode = config.get('core', {}).get('debug', False)
 
-    channels = {
+    plugin_kwargs = {
         'success_channel': asyncio.Queue(),
         'error_channel': asyncio.Queue(),
     }
-    plugin_names, plugins, errors, metrics = plugins_loader.load_plugins(
-        config, channels)
+    plugin_names, plugins, errors, plugin_kwargs = plugins_loader.load_plugins(
+        config, plugin_kwargs)
     if errors:
         for err_plugin, exc in errors:
             base_msg = 'Plugin was not loaded: {err_plugin}'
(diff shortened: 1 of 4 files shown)


tests/unit/test_main.py
```diff
@@ -239,7 +239,7 @@ def event_loop_mock(mocker, monkeypatch):
 @pytest.mark.parametrize('has_active_plugins,exp_log_count,errors', (
     (True, 2, []),
     (True, 2, [('not_a.plugin', conftest.plugin_exc_mock())]),
-    (False, 2, []),
+    (False, 1, []),
```
econchick: I think this originally passed because of an error message before this was fixed.

gordon/main.py
```diff
@@ -195,7 +195,7 @@ def run(config_root):
 
     route_config = config.get('core', {}).get('route', {})
     msg_router = _setup_router(
-        route_config, message_handlers, metrics, **channels)
+        route_config, message_handlers, **plugin_kwargs)
```
econchick: In plugin_loader.py, we're editing this dict value. Adding a `plugin_kwargs.copy()` adds a bit of overhead, although negligible. I figured since we're setting plugin kwargs to plugins and not separating out the `metrics` object, that we could keep the same convention here. Open to other ideas though!

descent3me: Ah, good catch! I didn't know plugin_kwargs was used outside of the plugin loader, but I guess that's just because of the different names. There's some argument to be made for not editing shared resources inside of a function, since it's more opaque. *If* you wanted to do it a different way, you could return `plugin_kwargs` from `load_plugins` instead of `metrics`.  That might make it more obvious that it has different values than the dict that went into the function.

econchick: @descent3me I like that approach - updated!

descent3me:
lgtm +1

PR #35, opened 2018-06-19
Improve async handling of polling channel

econchick:
Previously, while the tasks w/i the _poll_channel are async, each loop within the method would tend to work thru each message one at a time (seen by the messages inflight gauge fluctuating between 0 and 1). This patch allows the multiple _poll_channel task at once, managing many more messages in flight than just 1.

(copy-pasta of commit msg)
Previously, while the tasks w/i the _poll_channel are async, each loop within the method would work thru each message one at a time (seen by the messages inflight gauge fluctuating between 0 and 1). This patch allows the multiple _poll_channel task at once, managing many more messages in flight than just 1.

N/A

I'll cut a new release once approved.

**Release note**:

```release-note
Fix routing for handling more than one message at a time.
```

cc @spotify/alf

### DIFF
--- a/gordon/router.py
+++ b/gordon/router.py
@@ -170,7 +170,7 @@ class GordonRouter:
         return True
 
     async def _poll_channel(self):
-        while True:
+        if not self.success_channel.empty():
             event_msg = await self.success_channel.get()
 
             # graceful shutdown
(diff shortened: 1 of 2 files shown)


descent3me:
Seems ok to me 👍

PR #38, opened 2018-06-20
Respect ulogger configuration

econchick:

The `ulogger` defaults for syslog, stream, and stackdriver may not be good for `gordon` users. This patch passes the logging configuration down to `ulogger` for handling so we can override defaults.

Fixes #

**Release note**:

```release-note
Add support for ulogger configuration
```

cc @spotify/alf

### DIFF
--- a/gordon.toml.example
+++ b/gordon.toml.example
@@ -12,6 +12,10 @@ publish = "cleanup"
 [core.logging]
 level = "info"
 handlers = ["syslog"]
+format = "%(created)f %(levelno)d %(message)s"
+date_format = "%Y-%m-%dT%H:%M:%S"
+address = ["10.99.0.1", "514"]
+
 
 # Plugin Config
 ["foo"]
--- a/gordon/main.py
+++ b/gordon/main.py
@@ -95,12 +95,14 @@ def setup(config_root=''):
     """
     config = _load_config(root=config_root)
 
-    logging_config = config.get('core', {}).get('logging', {})
-    log_level = logging_config.get('level', 'INFO').upper()
-    log_handlers = logging_config.get('handlers') or ['syslog']
+    logging_config = config.get('core', {}).get('logging', {}).copy()
+
+    log_level = logging_config.pop('level', 'INFO').upper()
+    log_handlers = logging_config.pop('handlers', ['syslog'])
 
     ulogger.setup_logging(
-        progname='gordon', level=log_level, handlers=log_handlers)
+        progname='gordon', level=log_level, handlers=log_handlers,
+        **logging_config)
 
     return config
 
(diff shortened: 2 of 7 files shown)


gordon/main.py
```diff
@@ -99,8 +99,13 @@ def setup(config_root=''):
     log_level = logging_config.get('level', 'INFO').upper()
     log_handlers = logging_config.get('handlers') or ['syslog']
```
descent3me: You could do the `pop`s already up here and save a couple lines.

gordon.toml.example
```diff
@@ -12,6 +12,10 @@ publish = "cleanup"
 [core.logging]
 level = "info"
 handlers = ["syslog"]
+format = "%(created)f %(levelno)d %(message)s"
+date_format = "%Y-%m-%dT%H:%M:%S"
+address = ["10.99.0.1", "514"]
```
descent3me: I'm not sure it's a problem (because [SysLogHandler doesn't specifically mention it will fail with a string port](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.SysLogHandler), but it looks like ulogger [won't cast the second value to an int](https://github.com/spotify/ulogger/blob/master/ulogger/syslog.py#L81).

gordon/main.py
econchick: I thought about that, but I don't want to globally edit config. IIRC the metrics logger uses the level up here if it isn't set in its own config.

gordon.toml.example
econchick: dang, my eyes went to [this](https://github.com/spotify/ulogger/blob/master/ulogger/syslog.py#L94) thinking it did. hm I wonder if there's a graceful way to handle this in gordon. Trying to avoid fixing ulogger but I think that's probably what I should do...

gordon/main.py
descent3me: Sorry, I meant you could do the `copy` up here too, as part of line 98 maybe?  But if you think it makes things more confusing, I have no problem skipping it.

econchick: ahhh I get what you're saying; yeah that'd look nicer, good idea.

econchick: hm I don't think I can quickly fix this up because of the ability to set a default in `.get` 🤔

gordon.toml.example
econchick: @descent3me mind taking a look at this? it's just a quick+dirty fix: https://github.com/spotify/ulogger/pull/1

gordon/main.py
descent3me: I think `.pop` takes a default as well.

econchick: wat. I tried earlier and it wasn't behaving. (╯°□°）╯︵ ┻━┻

econchick:
ho'boy. I think this will do it 🤞 pardon my tripping up @descent3me. mind taking another looksee?

descent3me:
lgtm!  👍

PR #41, opened 2018-07-26
[ffwd] Don't mutate the context passed to metrics

descent3me:
It's reasonable that context may be reused for multiple calls to metrics plugins, so we should not be mutating the object that comes in.

Fixes #

@spotify/alf 

**Release note**:

```release-note
* Remove mutation of the context passed to ffwd plugin.
```


ocsw:
+1 Would it be clearer to use copy.deepcopy though?  Or maybe add a comment to make it clear why you're doing it this way?

descent3me:
Since what we want is a different dictionary anyways with some overlapping key-value pairs, I think that's essentially what this code shows.  I'm not sure that a copy would be any clearer.

